PR #9, opened 2020-01-17
average true range

DaveSkender:
add Average True Range indicator [AB#522](https://dev.azure.com/skender/5123ca47-74f2-4d67-a5d4-c4d90b8d670a/_workitems/edit/522)

### DIFF
--- a/Indicators/AvgTrueRange/README.md
+++ b/Indicators/AvgTrueRange/README.md
@@ -12,7 +12,7 @@ IEnumerable<AtrResult> results = Indicator.GetAtr(history, lookbackPeriod);
 
 | name | type | notes
 | -- |-- |--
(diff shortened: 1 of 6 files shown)


Indicators/AvgTrueRange/README.md
```diff
@@ -0,0 +1,52 @@
...
+``` C#
+// usage
+IEnumerable<AtrResult> results = Indicator.GetAtr(history, lookbackPeriod);  
+```
+
+## Parameters
+
+| name | type | notes
+| -- |-- |--
+| `history` | IEnumerable\<[Quote](/GUIDE.md#Quote)\> | Historical Quotes data should be at any consistent frequency (day, hour, minute, etc).  You must supply at least 2×`N` periods of `history`.  Since this uses a smoothing technique, we recommend you use at least 250 data points prior to the intended usage date for maximum precision.
```
DaveSkender: This is incorrect

Issue #22, opened 2020-05-21
Get MACD


GelLiNN:
I'm getting my historical data from yahoo finance, I have ensured everything is present in the Quote class, and then I pass the data into

`
int fastPeriod = 12;
int slowPeriod = 26;
int signalPeriod = 9;
IEnumerable<MacdResult> results = Indicator.GetMacd(history, fastPeriod, slowPeriod, signalPeriod);
`
But it gives me an exception on this line, every time saying "No historical quote provided".

I also have a stack trace for you if that's helpful:

`at Skender.Stock.Indicators.Cleaners.PrepareHistory(IEnumerable`1 history)\r\n   at Skender.Stock.Indicators.Indicator.GetEma(IEnumerable`1 history, Int32 lookbackPeriod)\r\n   at Skender.Stock.Indicators.Indicator.GetMacd(IEnumerable`1 history, Int32 fastPeriod, Int32 slowPeriod, Int32 signalPeriod)\r\n   at PT.Middleware.Indicators.GetIndicatorComposite(String symbol, String function, IEnumerable`1 history, Int32 daysToCalculate) in C:\\Users\\linki\\Dev\\Pro-Trades\\Middleware\\Indicators.cs:line 55`

I tried preparing it with the Cleaners class but it is still giving me the same exception.  Please tell me what the fix is or otherwise fix this bug?

Thanks a bunch!

DaveSkender:
[Unit test shows it works](https://dev.azure.com/skender/Stock.Indicators/_build/results?buildId=967&view=ms.vss-test-web.build-test-results-tab&runId=1000228&resultId=100018&paneView=debug), so the exception is catching something with your data, which is intended behavior, to help you troubleshoot.

``` csharp
            if (history == null || !history.Any())
            {
                throw new BadHistoryException("No historical quote provided.");
            }
```

GelLiNN:
I figured out what was going wrong.  I was providing only 7 days of historical data, when the MACD indicator needs much more than that to compute effectively.  I expanded it out to the last 100 days and it seems to be computing correctly.  Thank you!

DaveSkender:
Ah, yes.  You'd have to provide at least 2×`slowPeriod` amount of data to get any results.  And since MACD uses a smoothing technique, we recommend at least 200-250 extra data points for maximum precision.  I'm taking a note to improve documentation and to perhaps add some additional indicator specific exceptions.

PR #24, opened 2020-05-22
chore: add insufficient history exceptions

DaveSkender:
This is a helper exception that let's you know if you're not providing sufficient history.  #22

### DIFF
--- a/Indicators/Ulcer/Ulcer.cs
+++ b/Indicators/Ulcer/Ulcer.cs
@@ -13,6 +13,15 @@ namespace Skender.Stock.Indicators
             // clean quotes
             history = Cleaners.PrepareHistory(history);
 
+            // exceptions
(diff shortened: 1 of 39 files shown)


Indicators/Ulcer/Ulcer.cs
```diff
@@ -13,6 +13,15 @@ public static IEnumerable<UlcerIndexResult> GetUlcerIndex(IEnumerable<Quote> his
             // clean quotes
             history = Cleaners.PrepareHistory(history);
 
+            // exceptions
+            int qtyHistory = history.Count();
+            int minHistory = lookbackPeriod;
+            if (qtyHistory < minHistory)
+            {
+                throw new BadHistoryException("Insufficient history provided for Beta.  " +
```
DaveSkender: Ulcer Index

PR #86, opened 2020-08-15
setup CI with Azure Pipelines with YAML

DaveSkender:


DaveSkender:
removing to recreate the PR, to test that it works properly

PR #113, opened 2020-09-07
Changed Volume long to decimal

baidenx7:
Indicators.Volume type changed
ObvResult would have needed to be changed to match, didn't want to take that step since it was oiut of the scope of the issue so I cast Volume back to long

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -57,7 +57,7 @@ Historical quotes should be of consistent time frequency (e.g. per minute, hour,
 | `High` | decimal | High price
 | `Low` | decimal | Low price
 | `Close` | decimal | Close price
-| `Volume` | long | Volume
+| `Volume` | decimal | Volume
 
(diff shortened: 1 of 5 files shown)


DaveSkender:
Thanks for helping!  I think it's okay to update OBV as well, no point in double conversions of data types.  I expected some regressive impact where Volume is used.

DaveSkender (changes requested):
A few more updates needed:

- [ ] Update `GUIDE.md` in root folder to reflect change in `Volume` data type for `Quote` class
- [ ] Change OBV result data type from `long` to `decimal` (in both indicator code and in `ObvResult` class)
- [ ] Update OBV documentation `README.md` to reflect changed data type
- [ ] Remove `(long)` cast for `Volume` in IndicatorTests HistoryData files

DaveSkender (approved):
Excellent!  Thank you and congratulations on your first open-source contribution!

PR #127, opened 2020-09-13
update build definition

DaveSkender:


DaveSkender:
Was unable to get rid of Warning in build

Issue #170, opened 2020-11-10
Cannot reuse same history list with different slices on different indicators


MithrilMan:
Hello
I'm trying your library (thanks for the effort so far) and I've found a bug regarding the usage of the quotes to generate different indicators using the same set of historical quotes but taking a different slice of them

Debug was hard on my side, so I took a look at your code and I see that your model classes have an internal Index that you use and I think that's the problem.

Suppose I get historical data from my POCO classes and convert them to your `Quote` model
suppose I've a list `List<Quote> history` of 100 `Quote`s

Now if I call
```
Indicator.GetBollingerBands(history.TakeLast(20), 20);
Indicator.GetRsi(history.TakeLast(14), 14);
```

I get a
`System.ArgumentOutOfRangeException: 'Index was out of range. Must be non-negative and less than the size of the collection.'`

I think the problem is that if I reuse Quote objects taking a different slice of the array, the bugs pops out because there is the Index field inconsistence inside your model classes

This means that historical quotes cannot be reused this way but I must always use the same set and I think it's not good because of performance (I'm not interested in applying indicators on the whole historical set in my scenario

Thanks

MithrilMan:
here a screenshot to show you current data that gets passed to the second indicator
as you see, first item starts from Index 7 (and I don't have a way to reset these indexes I think)

![image](https://user-images.githubusercontent.com/5107375/98615203-59e0b880-22fa-11eb-97b2-493fcd111501.png)

DaveSkender:
Yah, the Index number itself is important for locating the start of the dataset relative to the lookback period.  Using TakeLast cuts it off unnaturally.  Since history was designed for reusability, it doesn’t recompose the Index (for performance reasons).

For now, you might try to cut the results down instead of the input history.

Let me think about it.  There’s likely both another workaround, so you can cut it like you depict, and perhaps a bug fix where we detect bad indexes and do a recompose.  I’ll also consider options for dropping the Index altogether as well — can probably find a better way.

MithrilMan:
> Let me think about it. There’s likely both another workaround, so you can cut it like you depict, and perhaps a bug fix where we detect bad indexes and do a recompose. I’ll also consider options for dropping the Index altogether as well — can probably find a better way.

An easy way before implementing a proper solution could be to implement an extension of `IEnumerable<Quote>` in your library that reset the indexes, so external library can call it

that way I can use that extension like `history.Reset()` or something like that

DaveSkender:
[...]

PR #171, opened 2020-11-10
check first Index value

DaveSkender:
Fix for reported bug #170.  A library internal error occurs when a user modifies `history` length after it was previously used or cleaned.  The fix allows the internal cleaner to recompose an internal Index if it detects a bad starting value, which can occur if you decide to reuse and cut down `history` with methods such as `TakeLast`.

I'm also adding an *undocumented* `history.RemoveIndex()` extension that will allow a user to remove the internal index from a previously utilized or cleaned `history`.  With this fix, you would not need this extension; however, I'm adding it in case there are unidentified reasons for a user to need to remove the internal indexing.

### DIFF
--- a/indicators/_Common/Cleaners.cs
+++ b/indicators/_Common/Cleaners.cs
@@ -49,6 +49,24 @@ namespace Skender.Stock.Indicators
         }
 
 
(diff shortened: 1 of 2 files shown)


DaveSkender:
- [ ] add unit tests

Issue #179, opened 2020-11-15
Implement IQuote interface to expand utility of quote class


MithrilMan:
I'm not going to criticize you work at all, it's cool that you put lot of effort in this project, I'm just offering some help if you want to talk about the design of the library and possible improvements both in memory and performance.

An example is about your Quote class, I think it's not a good idea to require a specific class because that way an user has to convert it's own models to Quote with problems both at performance level and design

A proper solution would be to use a IQuote interface, so that it's easy to adapt our own model because it's a matter of implementing that interface in our model without having to convert any object (most of the time our source model already contains close, high, low, date etc... so it's a no brainer to adapt)

Of course Quote class can still be there in the library for those that want to use that model.

Also the Index property is something not needed as I pointed out in a comment here https://github.com/DaveSkender/Stock.Indicators/issues/170#issuecomment-727620632

but even if needed, having a different interface like IIndexedQuote required only by indicators that need to manage an index would be enough (and if we want to adapt our model adding the index property, we can do that with explicit interface declaration so that our model will not be polluted externally)

If you are open to changes I can try to find some time to contribute and do some benchmark too

DaveSkender:
Hey, I'm always looking for passionate people who want to contribute, either with feedback or more directly by helping, so, thank you for that.  I think you have some good ideas here, worth exploring.

DaveSkender:
One of the key design decisions to keep in mind is that the `Quote` is the natural shape of historical data from pretty much all data feeds and sources; as such, I do not want to force users (developers) to do a lot of manipulation of data before using it in the library.  If I wanted to spend time coding data transformation and matching and rematching data after I get results, I'd use some of the other libraries out there, like Tulip or TA-Lib, which are very difficult to use IMHO.

MithrilMan:
Sure, my main problem is time but I'll fork the repo and implement a couple of Indicators following my proposal to showcase
I'll keep you updated on this issue

DaveSkender:
Thank you.  In the meantime, I'll take a look at removing the internal Index altogether.  It's something I've been interested in doing anyway.

MithrilMan:
removing index will be really appreciated and will even solve the issue I opened from the root (about trimming date).

I think that removing index will mean even that you must expect data passed is already sorted by date and don't contains date duplication, but I think it's good because it has performance implication to do that inside your indicators
[...]

PR #190, opened 2020-11-21
add IQuote generic interface

MithrilMan:
all indicators makes use of generics and IQuote
implements #179 

Tests aren't affected because works as they were out of the box

Documentation need to be updated

With this change we are no forced to use the built in `Quote` type so we can use our own models (that have to implement the IQuote interface)

performance wise is good to have to cast or convert objects to use indicators (this mean that even memory allocation is saved in such scenarios)

### DIFF
--- a/indicators/Adl/Adl.cs
+++ b/indicators/Adl/Adl.cs
@@ -7,11 +7,11 @@ namespace Skender.Stock.Indicators
     public static partial class Indicator
     {
         // ACCUMULATION / DISTRIBUTION LINE
-        public static IEnumerable<AdlResult> GetAdl(IEnumerable<Quote> history, int? smaPeriod = null)
+        public static IEnumerable<AdlResult> GetAdl<TQuote>(IEnumerable<TQuote> history, int? smaPeriod = null) where TQuote : IQuote
         {
 
             // clean quotes
-            List<Quote> historyList = history.Sort();
+            List<TQuote> historyList = history.Sort();
 
             // check parameters
             ValidateAdl(history, smaPeriod);
@@ -23,7 +23,7 @@ namespace Skender.Stock.Indicators
             // get money flow multiplier
             for (int i = 0; i < historyList.Count; i++)
             {
-                Quote h = historyList[i];
+                TQuote h = historyList[i];
                 int index = i + 1;
 
                 decimal mfm = (h.High == h.Low) ? 0 : ((h.Close - h.Low) - (h.High - h.Close)) / (h.High - h.Low);
@@ -58,7 +58,7 @@ namespace Skender.Stock.Indicators
         }
 
 
-        private static void ValidateAdl(IEnumerable<Quote> history, int? smaPeriod)
+        private static void ValidateAdl<TQuote>(IEnumerable<TQuote> history, int? smaPeriod) where TQuote : IQuote
         {
 
             // check parameters
--- a/docs/GUIDE.md
+++ b/docs/GUIDE.md
@@ -5,9 +5,9 @@
 - [Prerequisite data](#prerequisite-data)
 - [Example usage](#example-usage)
 - [About historical quotes](#quote)
-- [Using custom Quote classes](#using-custom-quote-classes)
+- [Using custom quote classes](#using-custom-quote-classes)
 - [Validating historical quotes](#validating-historical-quotes)
-- [Using derived Results classes](#using-derived-results-classes)
+- [Using derived results classes](#using-derived-results-classes)
 - [Generating indicator of indicators](#generating-indicator-of-indicators)
 - [Contributing guidelines](CONTRIBUTING.md)
 
@@ -46,7 +46,7 @@ Console.WriteLine("SMA on {0} was ${1}", result.Date, result.Sma);
 SMA on 12/31/2018 was $251.86
 ```
 
-See [using custom Quote classes](#using-custom-quote-classes) if you prefer to use your own quote class.
+See [using custom quote classes](#using-custom-quote-classes) if you prefer to use your own quote class.
 
 See [individual indicator pages](INDICATORS.md) for specific guidance.
 
@@ -75,7 +75,7 @@ Note that some indicators, especially those that are derived from [Exponential M
 
 For example, if you are using daily data and want one year of precise EMA(250) data, you need to provide 3 years of total historical quotes (1 extra year for the lookback period and 1 extra year for convergence); thereafter, you would discard or not use the first two years of results.
 
-## Using custom Quote classes
+## Using custom quote classes
 
 If you would like to use your own custom `MyCustomQuote` _quote_ class, to avoid needing to transpose into the library `Quote` class, you only need to add the `IQuote` interface.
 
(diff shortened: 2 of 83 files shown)


DaveSkender:
I'll take a look at this a bit more later today.  Generally speaking, this idea looks good and will be something that we can accept after we get through the code review.  @MithrilMan are you at all interested in updating the docs as well; if not, I'll take care of it.  Just checking to see if you intended to do it.

indicators/Adl/Adl.cs
```diff
@@ -7,11 +7,11 @@ namespace Skender.Stock.Indicators
     public static partial class Indicator
     {
         // ACCUMULATION / DISTRIBUTION LINE
```
DaveSkender, line 9: I'm not an expert on Interface patterns, so I might need some help understanding.  Why use `TQuote` instead of a simpler use of `IQuote`:

```
        // ACCUMULATION / DISTRIBUTION LINE
        public static IEnumerable<AdlResult> GetAdl(IEnumerable<IQuote> history, int? smaPeriod = null)
        {

            // clean quotes
            List<IQuote> historyList = history.Sort();

            // check parameters
            ValidateAdl(history, smaPeriod);

            // initialize
            List<AdlResult> results = new List<AdlResult>();
            decimal prevAdl = 0;

            // get money flow multiplier
            for (int i = 0; i < historyList.Count; i++)
            {
                IQuote h = historyList[i];
                [..]
```

MithrilMan: it's because covariance and contravariance
we are not passing a IQuote here but an `IEnumeration<IQuote>` and there are limits because of that
this post may contains useful information to you https://stackoverflow.com/questions/2033912/c-sharp-variance-problem-assigning-listderived-as-listbase
there are places that would require you to explicitly cast from a list of interfaces to a list of your concrete types (`ToList<YourClassThatImplementIQuote>`)

Using directly IQuote was my first attempt that I had in this branch, you can see the diffs
https://github.com/DaveSkender/Stock.Indicators/compare/master...MithrilMan:quote-to-iquote?expand=1

you can see that this way you are forced to work with IEnumerable<IQuote> while with generics you have typed list that uses the type of your IQuote implementation, so this solves the problem without requiring casts


About the documentation, I didn't have much time right now so I committed the most important pieces (and because of time constraint I didn't wanted to waste time on something that wasn't sure to be accepted/merged)

DaveSkender: Yah, it’s good to hold on docs.  I’m still trying to wrap my head around it.

MithrilMan: About the use of generics?
Generics are quite powerful and a core feature on c# and allows you to work with Typed objects.

If you want instead an usecase about why the interface is useful, imagine having a feed of data coming from a 3rd party library (that have it's own model) then you are building a system where you have your own model because you can obtain data from different feeds for example, or because (it's always a good choice) you don't want to depend on 3rd party library models in case you want to swap them later, or provide alternative implementations for that part of software.
Now what you do is mapping the feed models into your own models (and if the 3rd party library doesn't allow you to specify your own model as a type you are forced to do the mapping.
Now it's your library turn, the user wants to use it to generate indicators on the data he has (in it's own models).
With current implementation he's forced to do another mapping into your Quote object, while with this modification, he has just to implement the IQuote interface on its own data model to map its own field into your required fields

Now he's able to feed, let say a `List<MyDataCandleStick>` into your indicator (where `MyDataCandleStick` is implementing IQuote) without having to spend time mapping from one type to another (less code, less computation, less memory allocation)

In a real time application, where data is feeded continuously (e.g. from a websocket that stream data), not having to perform all the mapping everytime is important (the alternative would be to keep a copy (so waste of memory) of the stream as a `IEnumerable<Quote>` or similar and just map the fresh data and append it to historical (like I'm actually doing) but it's far from efficient and force you to maintain more code too

For a simple usage of the library, nothing change for your users, they can use your Quote class as they are doing now, for advanced useage you allow programmers to reuse their model (just by implementing the interface)

DaveSkender: I get the interface part, a user being able to keep and use their model.  I'm mostly struggling with generic `TQuote` vs `IQuote` usage here and why the simpler `IQuote` interface won't work (or is somehow inferior).  `TQuote` _appears_ to add complexity and I just want to make sure it's adding enough value, and to personally  learn a bit more about it.

DaveSkender:
I've made an internal pre-release package for this version and am dogfooding it in my personal trading system.
You can get this package [here](https://dev.azure.com/skender/Stock.Indicators/_packaging?_a=package&feed=Internal&package=Skender.Stock.Indicators&protocolType=NuGet&version=1.1.12-PullRequest0190.4).

DaveSkender:
@MithrilMan if you have time, please also review docs to see if they make sense.  I've taken a shot at updating these accordingly.

MithrilMan:
Ok I will, btw I think we could change the interface to only require getters and not setters because if I recall, IQuote is only read (there are few cases where you create temporary Quotes to call indexers from within another indexer but there you still use Quote that can still set values of course)

Having a simpler interface requires less work to adapt custom model (no need to define setters) and so if a user has for example already a `MyClose` property that represents the IQuote Close property in its model, the required implementation would just be
`public decimal Close => MyClose`

DaveSkender:
> Probably I'll rewrite the "using your custom model" section too in the documentation

Please do.  Sounds like you can provide a more thoughtful example.

MithrilMan:
a note about the section "Using derived classes"
There is absolutly no point deriving result like in your example because the fact that you inherit from EmaResult is just to not have to declare the properties you have already in EmaResult, explicitly into MyEma.
But performance wise what you do it's a copy of all properties from a type to another and if you type explicitly the properties of EmaResult in your MyEma it would work the same.

Since your library doesn't allow you to pass your custom result type (that would instead be helpful), what I suggest you to write in that section is a composition pattern, instead of inherit EmaResult, just wrap it as a property

```
// your custom derived class
public class MyEma
{
  public int MyId { get; set; }
  public EmaResult Result {get; set}
}
```
This way you don't need to deep copy all the properties using the Select but just populate 2 properties: the additional field "MyId" and the result (this would mean performance improvement too)

MithrilMan:
Also note that you can take advantage of Generics to do a class like
// your custom derived class
public class MyIndicatorResult<TResult>
{
  public int MyId { get; set; }
  public TResult Result {get; set}
}
```

so you can reuse the same class to inject a custom Id in every indicator result

MithrilMan:
ok updated documentation and interface definition

DaveSkender:
> There is absolutly no point deriving result like in your example because the fact that you inherit from EmaResult is just to not have to declare the properties you have already in EmaResult, explicitly into MyEma.

I think this is a personal stylistic choice here.  Some people may not like nested classes.  If you want to add an alternative "or" example with the nested result, I'm okay with that.  Generally, though, we should discuss generic result classes under another issue.

MithrilMan:
It's not just stylistic because if you have a resul class that have 20 fields, you'll have to allocate memory for these 20 fields and you have to perform the copy of these values (or their pointers) while having a reference on the original result prevent that
And if you are working with multiple indicators and you want just to carry your additional ID for every indicator, following your approach you are forced to have a different type for each indicator you are working with, while with the generic approach (my last example) you don't.

Since it's not an implementation detail of your library but an usage suggestion, a good programmer should be able to implement the integration at its best (but giving an advanced advice could be useful also).
Anyway I agree that this should be on another issue.

DaveSkender (dismissed):
I think this looks pretty good at this point.  I'll let this sit for a few hours for additional comments then merge it later today.

DaveSkender (approved):

PR #191, opened 2020-11-21
chore: update docs

DaveSkender:
Adding Excel files for manual calculations, for recordkeeping (some were not kept and are missing) + minor fixes to docs.


DaveSkender:
- [ ] fix background color on Williams %R chart

Issue #342, opened 2021-02-03
Fractal Chaos Bands


DaveSkender:
Add [Fractal Chaos Bands](https://library.tradingtechnologies.com/trade/chrt-ti-fractal-chaos-bands.html), requested by @metathron.

DaveSkender:
This is now available in version 1.8.0

Issue #198, opened 2020-12-05
Add Williams Fractal Indicator


DaveSkender:
Request to add [Williams Fractal Indicator](https://www.investopedia.com/terms/f/fractal.asp) from @abcdamir2

DaveSkender:
This was implemented in version 1.3.0

PR #348, opened 2021-02-07
add Fractal Chaos Bands

DaveSkender:
- parameterize Williams Fractal evaluation window #198
- add Fractal Chaos Bands (FCB) #342


DaveSkender:
Does this look right?
image

Issue #318, opened 2021-01-13
SuperTrends is not correct?


l150039:
Hi team.
I find SuperTrends result is not correct.
I tested it by "bitcoin 1 day candle 2017-2020" and compared with tradingview indicator result but not matched. (bollingerbands result is correct! good!)

can you check? thanks!

DaveSkender:
Can you provide more information?

1. Link to the TradingView indicator
2. How are you using the library?  How much history, parameters, etc.

I’ll check it out.  As a side note, TradingView often uses user provided formulas, so it’s possible their formula is incorrect or that they made different assumptions.  Also, make sure you are using more than enough history, this indicator does have a warmup period where early vale’s in the time series will be less precise and should not be used.

DaveSkender:
On my initial review of [the code](https://github.com/DaveSkender/Stock.Indicators/blob/master/indicators/SuperTrend/SuperTrend.cs), I'm not able to identify issues.  Currently, I do not have enough information to determine if there's an error in this library, the TradingView indicator, or if the comparison is based on the same historical quotes.

If anyone can provide information to clearly show an error in this library's formula, please do.  Otherwise, I'm going to assume that the TradingView implementation is errant or using "wicks" as the breakpoint, or that the provided historical quotes data is somehow different.

I'll leave this issue open for a while to elicit more feedback; however, will close it after a few days if we can't find a problem.

l150039:
Hi Dave.

I test binance BTC-USDT (1day candle) Data. I will give you sample.
[BTCUSDT_1d_1610582399999_1610550000000.json.txt](https://github.com/DaveSkender/Stock.Indicators/files/5819925/BTCUSDT_1d_1610582399999_1610550000000.json.txt)

1. parsing data and try Indicator.GetSuperTrend.
(lookbackPeriod_supertrends = 10, multiplier_supertrends = 3)

2. and compare with binance tradingView (https://www.binance.com/en/trade/BTC_USDT)
![total](https://user-images.githubusercontent.com/77389426/104714105-22621280-5768-11eb-8120-49a9de3072f7.jpg)


Here is my result
![RES](https://user-images.githubusercontent.com/77389426/104714946-2e020900-5769-11eb-93b5-34649611e642.JPG)

and Binance sample
![1](https://user-images.githubusercontent.com/77389426/104714959-33f7ea00-5769-11eb-963b-8dc13dfc7116.jpg)
[...]

PR #384, opened 2021-03-01
Cci calculation error fixed.

SnnK:
Compared and verified on TradingView (BINANCE:BTCUSDT 4h)

### DIFF
--- a/indicators/Cci/Cci.cs
+++ b/indicators/Cci/Cci.cs
@@ -8,13 +8,13 @@ namespace Skender.Stock.Indicators
     {
         // COMMODITY CHANNEL INDEX
         /// <include file='./info.xml' path='indicator/*' />
-        /// 
-        public static IEnumerable<CciResult> GetCci<TQuote>(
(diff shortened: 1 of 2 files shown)


DaveSkender:
Can you describe the original error?  Also, please do not use TradingView as a benchmark or source of accuracy.  Most of their indicators are user provided and untested, see #318 as an example.

DaveSkender:
From what I can tell from your code, all you’ve done is replace True Price (correct formula) with close price (incorrect formula).

SnnK:
Hi @DaveSkender 

First of all thank you for this great repository.

I checked again.

An incorrect formula misled me. When it matched the TradingView results with my results, I didn't think much about it. Sorry.

DaveSkender:
Not a problem and thank you for your feedback.  Being able to periodically double-check our formulas is a good thing.  There's a lot of bad information out there and part of my job is to sift through it and ensure that we have the most reliable library as possible.

Issue #394, opened 2021-03-20
Python wrapper


DaveSkender:
Add Python wrapper and publish as a community Wheel package to PyPI.
We're seeking Python contributors to lead an effort to build a version-synced parallel path Python wrapper for this library.

## Pull-Requests To Do

- [x] Add baseline wrapper with 1-2 indicators to prove our approach - @LeeDongGeon1996
- [x] Add initial unit tests to verify the implementation - @LeeDongGeon1996
- [x] Add packaging and unit testing to the build - @LeeDongGeon1996
- [ ] Add remaining indicators and unit tests - @LeeDongGeon1996
- [ ] Final sanity check and any needed refactoring
- [ ] Add general "Python support" documentation to expand the Guide (e.g. how to use the PyPI package)
- [ ] Update individual indicator documentation for Python usage
- [ ] Add deployment to package manager(s), deploy

DaveSkender:
@LeeDongGeon1996 if you or anyone is interested in [contributing](https://daveskender.github.io/Stock.Indicators/docs/CONTRIBUTING.html) to this, let us know and we can start to discuss implementation details.  I haven't done a lot of research, so I'm unclear on whether we need to add any more C# code to make this easier.  I'm definitely here to help and contribute in those areas if needed.  I was thinking we put all the independent Python wrapper code in a new `/wraps/Python` folder in the repo.

My only ask is that we implement in small increments.  In other words, the first Pull Request should only include one or two indicators, without any documentation updates, and will be used to establish the baseline approach.

We can then figure out a strategy to get all the way to a published package that people can get from the public Python package managers.  I'll start a Pull Request To-Do list above to capture the sequence of increments and who's working on what parts to coordinate amongst collaborators.  Just announce here what piece you're picking up and I'll put your name next to it, then check it off when done.  Feedback appreciated.

LeeDongGeon1996:
Hi, first of all, I'm happy to get a chance to contribute this project.
There are some topics I'd like to talk about:

 * **Why don't you separate the repo for C# and Python?**
   I think some of users using Python may want to look inside the project(library), but there is high probability of that they are only familiar with Python code and they might want to read only Python not a C#. Of course, most of important core implementation is written in C# and people should read C# code if they want to know how the module works exactly. But the options always should be there at least in my thought. That's why I think separating repo makes it easier to explore the code by language. please let me know your thought about this.

 * **Need to convert primitive type of C# and base class like `Qoute`, FIRST**
[...]

PR #397, opened 2021-03-26
add baseline for Python wrapper

LeeDongGeon1996:
## Description

 * Add wrapper functions for DateTime, Decimal and List of C#.
 * Add wrapper for SMA indicator

Implements #394

Additional comment:
1. I did mistake. I wanted to make separate 2 PRs but it is created automatically as one PR.
2. I wrote comment and description for `cstypes`, but not `indicators` yet. But I've already finished testing with my `CustomQoute`(meaning [Using custom quote classes](https://daveskender.github.io/Stock.Indicators/docs/GUIDE.html#prerequisite-data) is also working.). I didn't include my testing code in this PR because it is not formatted and a bit of a mess. Maybe, we need to write better testing code later.

## Checklist

- [ ] I have made corresponding changes to the documentation, including INDICATORS.md, README.md, info.xml, etc
- [ ] I have added unit tests that prove my fix is effective or that my feature works, and achieves sufficient code coverage
- [ ] I have added or run the performance tests that depict execution times similar to others
- [ ] New and existing unit tests pass locally with my changes

## Acknowledgements


wraps/python/src/cstypes/Decimal.py
```diff
@@ -0,0 +1,23 @@
...
+   
+    >>> cs_decimal = Decimal(2.5)
+    >>> cs_decimal
+    2.5
+    """
+    
+    if not isinstance(decimal,(int, float)):
+        raise TypeError("Only int or float are allowed")
+
+    return CsDecimal(float(decimal))
```
DaveSkender: Are we going to lose precision with this?  It's probably not an issue if we're only dealing with inputs and outputs to the library, but want to make sure we're not creating an accuracy problem in the wrapper.

wraps/python/src/cstypes/__init__.py
```diff
@@ -0,0 +1,6 @@
+import clr
+clr.AddReference(r'dll/Skender.Stock.Indicators')
```
DaveSkender: Just an FYI, but we'll have to figure out how to automatically update this DLL when the C# project builds; or, reference the native location of the `bin\Release` folder for the output.  Only concerned that we're using the most appropriate DLL version here.

wraps/python/src/indicators/Qoute.py
```diff
@@ -0,0 +1,11 @@
+from Skender.Stock.Indicators import Quote
+from cstypes import DateTime, Decimal
+
+class Qoute(Quote):
```
DaveSkender: I realize you're probably spelling this differently to avoid naming conflicts, but `Qoute` it looks like a typo.  I noticed you were using `Cs` as a prefix in the `System` types, maybe we can do something like that here to make it look more intentional?

wraps/python/src/indicators/Qoute.py
```diff
@@ -0,0 +1,11 @@
...
+    def __init__(self, date, open = 0, high = 0, low = 0, close = 0, volume = 0):
```
DaveSkender: Are these defaults required?  I think `null` would be a better default on the C# handling side.  `0` is an actual viable price value and shouldn't be accidentally applied.

wraps/python/src/indicators/SMA.py
```diff
@@ -0,0 +1,37 @@
...
+    sma_extended_list = [ SmaExtendedResult(r) for r in sma_extended_list ]
+
+    return sma_extended_list
+
+class SmaResult:
+    def __init__(self,sma_result):
+        self.Date = sma_result.Date
+        self.Sma = to_pyfloat(sma_result.Sma)
+
+    def __str__(self):
```
DaveSkender: How is this string concatenation used?  Seems like an odd thing, but there's a lot I don't know about Python.

wraps/python/src/indicators/Qoute.py
DaveSkender: Maybe `... import Quote as CsQuote` then keep `Quote` as the native Python class name?

DaveSkender (changes requested):
I don't see any major issues here.  Layout seems appropriate, looks like you're following good general practices.  I've added some small comments and questions to review.

wraps/python/src/cstypes/__init__.py
DaveSkender: When it comes time to run the build, we'll do the C# build first, then somehow use that DLL when we package this.  I'd like to avoid manually copying the DLL to this location, if at all possible.

DaveSkender:
As a side tip, it helps to create a feature branch in your fork to make it easier to manage the separation of PRs.

wraps/python/src/cstypes/Decimal.py
LeeDongGeon1996: Actually, Seems like it's losing its precision. But I don't think it would be the problem at least in this function because `Decimal()` is used for history data(OHLCV), which is usually represented in two decimal points. 

But the problem is `to_pyfloat()`, used for converting `System.Decimal` from Result class. I found that Result Decimal lose its precision when it is converted by `to_pyfloat()`. 
(e.g. `0.0324479652254633451142381441` becomes `0.032447965225463345` )

The solution is using [`decimal.Decimal`](https://docs.python.org/3/library/decimal.html). It is C decimal arithmetic module in Python and It works without losing.

wraps/python/src/cstypes/__init__.py
LeeDongGeon1996: Yes, this is the thing I want to discuss.
I hope so, but still don't know where is the proper location for the dll. I place the dll in `.../wrap/python/dll` temporarily to develop first. 

Also, the path string for `AddReference()` have to be change by the directory where I run the code.  but I still cannot figure out where to be set. Because it is module and it should be able to be imported to any other directory where the users want.

Could you do some research for this? I couldn't find the solution until now.

wraps/python/src/indicators/Qoute.py
LeeDongGeon1996: Okay. That looks better for code consistency!

wraps/python/src/indicators/Qoute.py
LeeDongGeon1996: You're right. That will be better.
One question, is it null-acceptable in C# module?

wraps/python/src/indicators/SMA.py
LeeDongGeon1996: `__str__()` is used for type casting into string. Just for convenience.
Usually used for:
```python
>>> sma_list = Indicator.GetSma[Qoute]( some values... )
>>> print(sma_list[0])
2021-09-23    12.3
```
Without `__str__()`:
```python
>>> sma_list = Indicator.GetSma[Qoute]( some values... )
>>> print(sma_list[0])
<indicators.SMA.SmaResult object at 0x0A710700>
```

LeeDongGeon1996:
> As a side tip, it helps to create a feature branch in your fork to make it easier to manage the separation of PRs. Though, if you're only working on one at a time, it's not a problem.

Didn't know that. Thank you 😀

wraps/python/src/indicators/Qoute.py
DaveSkender: It will accept an "unset" value natively and it does translate to `null`.  However, I'm not 100% sure what happens when it is set as `null` and converted by the Python wrapper.  Let's give it a try and see what happens.

wraps/python/src/indicators/SMA.py
DaveSkender: Is that used for debugging or is it needed for the production code?  The reason I ask is if it's just debugging code, it probably shouldn't be in the final production code and may not be as efficient in higher-performing use cases.  What's your opinion?

wraps/python/src/cstypes/Decimal.py
DaveSkender: I think if we do some unit tests and values are still within significant digit accuracy, then it's not a problem.  We can revisit this later when we add unit tests if you don't think it will be a problem.

wraps/python/src/cstypes/__init__.py
DaveSkender: Are you able to use the .NET 5 version of the DLL?  If so, maybe we try `../../indicators/bin/Release/net5.0/` as the path, which will require a `dotnet build -c Release` when we're developing to produce it in its native location.  I'm hoping when we do the packaging, that relative path won't be an issue.  Let's try that and see what happens.

wraps/python/src/indicators/Qoute.py
DaveSkender: To be clear, the correct spelling is `Quote` and not `Qoute` in both cases, including the file name.

LeeDongGeon1996: Okay, then I'll set default to `None` which is null in Python. 
FYI, when C# module outputs `null`, Python accepts it as `None`.
Then, what we have to check is reverse case.

> To be clear, the correct spelling is Quote and not Qoute in both cases, including the file name.

Oh, sorry, I'll fix that.

wraps/python/src/indicators/SMA.py
LeeDongGeon1996: Ah..then, It's more like debugging code. `__str__()` is only needed for string expression of the object. In my opinion, removing from production code would be better.

wraps/python/src/cstypes/__init__.py
LeeDongGeon1996: I set `../../../indicators/bin/Release/net461/Skender.Stock.Indicators` as the path after doing `dotnet build -c Release`.

There is one problem.
When I import the dll in `net5.0`, `netcoreapp3.1` and `netstandard2.1`, it shows the error, "ModuleNotFoundError: No module named 'Skender'", maybe python cannot recognize the module.
But when I import the dll in `net461` and `netstandard2.0`, it works successfully.

Do you know the reason?

wraps/python/src/cstypes/Decimal.py
LeeDongGeon1996: I've tried in many ways, but couldn't find the solution for this. As I told you, OHLCV values usually are represented in up to two decimal-points so I think it is enough. BUT just in case as you worried, I tried to convert floating-number in Python into C# decimal without losing but Failed (in case of long float number like `0.0217609957385529775904996576`). 


I posted [the question](https://stackoverflow.com/questions/66841211/convert-python-floating-point-into-c-sharp-decimal) in Stack Overflow....
Anyway, we need to do unit test first.

wraps/python/src/cstypes/__init__.py
DaveSkender: I suspect this is a Python.NET compatibility issue where .NET Standard 2.0 is the best they can support.  Here's some discussion on the topic:
`https://github.com/pythonnet/pythonnet/issues/857`

This may be the solutions?
https://pypi.org/project/clr-loader

If we're stuck with the choice between .NET 4.6.1 and .NET Standard 2.0, I'd pick the latter as it's the more modern of the two.

wraps/python/src/cstypes/Decimal.py
LeeDongGeon1996: [[`107ef7d`](https://github.com/DaveSkender/Stock.Indicators/pull/397/commits/107ef7ddeeb39c4195ee2eb3241f58e514438568)] I got an answer from the posting, very simple.  I don't know why I was missing it. What was wrong with my brain🤐
Just use `Decimal.Parse()`:
```python
>>> from cstypes import Decimal
>>> print(to_pydecimal(Decimal("0.0234337688540165776476565071")))
0.0234337688540165776476565071

>>> from decimal import Decimal as PyDecimal
>>> print(Decimal(PyDecimal("0.0234337688540165776476565071")))
0.0234337688540165776476565071
```
 This can keep 28digit floating-point precision. Same with `System.Decimal` in C#

wraps/python/src/cstypes/__init__.py
LeeDongGeon1996: > but we'll have to figure out how to automatically update this DLL when the C# project builds;

I suddenly got a question about it. Did you mean auto-updating dll using network? The dll should be hard-copied at the end, I mean when we distribute this wrapper on PyPi.

Also understood that we should use NET Standard 2.0 at least now because of compatibility of Python .Net. But I couldn't find out what `clr-loader` is used for, for a few days. There is not enough description, just can guess on `https://github.com/pythonnet/pythonnet/issues/857`.

Anyway, there is another way we can set the path for dll on production code.
=> [2.6. Installing Package Data](https://docs.python.org/3/distutils/setupscript.html#installing-package-data)

Examples:
`https://stackoverflow.com/questions/57191483/include-a-dll-in-python-module-pypi`
`https://stackoverflow.com/questions/33216691/add-a-dll-so-to-a-python-built-distribution`

DaveSkender: 1. Version targeting: Ok, let's just use the .NET Standard 2.0 for now.  We can always revisit later when Python.NET upgrades itself to use newer .NET Core or .NET 5
2. DLL Loading: I was not thinking of the actual programatic loading, but glad you're thinking of that.  Mostly I wanted to make sure we're using the final version stamped DLL and not the one we might copy locally during development.  It's a chicken vs egg problem.  The final version stamped DLL is only produced during the [online Azure DevOps] build, so you'd never be able to copy it yourself beforehand.  I think what we can do is basically go back to your original dll folder, add the locally produced DLL manually, but then replace it with the stamped DLL during the final build (simple file copy command).
3. We'll discuss packaging in more detail later, but it's something to start thinking about.

LeeDongGeon1996: When I place the dll in `.../wraps/python/lib`, example is like the below:
```python
# cstypes/__init__.py
import os
import sys
import clr

dir = os.path.dirname(sys.modules["cstypes"].__file__)
path = os.path.join(dir, "../../lib/Skender.Stock.Indicators.dll")
clr.AddReference(path)

```

LeeDongGeon1996: Now, I get it!! I was totally misunderstanding your words. OMG.. I was thinking that you want to update automatically dll in distribution..ok, now, I know..

[[`fdc080b`](https://github.com/DaveSkender/Stock.Indicators/pull/397/commits/fdc080bb8452ca7c33ef0214f53d54497ff489ea)] By the way, the below solves the problem about relative path, I mentioned above.

When I place the dll in `.../wraps/python/lib`, example is like the below:
```python
# cstypes/__init__.py
import os
import sys
import clr

dir = os.path.dirname(sys.modules["cstypes"].__file__)
path = os.path.join(dir, "../../lib/Skender.Stock.Indicators.dll")
clr.AddReference(path)

```

DaveSkender (approved):
Looks good.  Anything else you want to do here, or are you ready to commit?

LeeDongGeon1996:
Nothing else. :)

DaveSkender:
@LeeDongGeon1996 Awesome, thank you.  After this merge, I'd recommend that you fully delete your Fork and recreate it; then start using feature branches to do the work.  This will help you refresh with `upstream` pulls and also to keep your fork's `master` branch from getting contaminated with dev commits.

Here's a good explanation of the workflow:
https://codeburst.io/a-step-by-step-guide-to-making-your-first-github-contribution-5302260a2940

LeeDongGeon1996:
Thank you for merging and helpful article! I'll follow it.

PR #398, opened 2021-03-27
add Efficiency Ratio to KAMA results

Fredrik-C:
I would find it usefull if KamaResult included Efficiency Ratio so that I can filter out busy/noisy periods where there is risk for false signals. 

- [x] I have made corresponding changes to the documentation, including INDICATORS.md, README.md, info.xml, etc
- [x] I have added unit tests that prove my fix is effective or that my feature works, and achieves sufficient code coverage
- [ ] I have added or run the performance tests that depict execution times similar to others
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/indicators/Kama/Kama.cs
+++ b/indicators/Kama/Kama.cs
@@ -60,12 +60,14 @@ namespace Skender.Stock.Indicators
                         // kama calculation
                         decimal? pk = results[i - 1].Kama;  // prior KAMA
                         r.Kama = pk + sc * sc * (h.Close - pk);
+                        r.ER = er;
                     }
 
                     // handle flatline case
                     else
                     {
                         r.Kama = h.Close;
+                        r.ER = 0;
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


indicators/Kama/Kama.cs
```diff
@@ -60,19 +60,22 @@ public static partial class Indicator
...
                         r.Kama = h.Close;
+                        r.ER = 0;
                     }
                 }
 
                 // initial value
                 else if (index == erPeriod)
                 {
                     r.Kama = h.Close;
+                    r.ER = 0;
```
DaveSkender: I think we should leave this as `null` or unset here instead of `0`.  A zero value would imply the last X periods were an extreme flatline case.

DaveSkender (changes requested):
This looks good.  I'll make a quick commit in your fork to remove the initial value, then merge.  Nothing more you need to do.  Thank you for your contribution!

DaveSkender (approved):

PR #424, opened 2021-05-03
add Python wrapper to build

LeeDongGeon1996:
## Description

Update `build.main.yml` to update the build on Azure Pipelines.

Implements #394 

## Checklist

- [ ] I have made corresponding changes to the documentation, including `INDICATORS.md`, `README.md`, `info.xml`, etc
- [ ] I have made corresponding changes to the `wraps` interoperability files
- [x] New and existing unit tests pass locally and in the build (below) with my changes

## Acknowledgements


DaveSkender:
Having fun with DevOps yet?  :smile:

LeeDongGeon1996:
Too much fun. 😂
Yes, I was watching the detail log on Azure Pipelines, but couldn't get enough information about it.
The problem is that ,on Windows, the test session is not terminated after all tests are passed. (On Linux, it shows segmentation error). I'm suspecting the part that reading directly from spreadsheet. When I skip that part, it works.

I read your C# test code and found that you create Quote data in hard-coded way, not fetching from spreadsheet. Is there any reason for this?

DaveSkender:
> I read your C# test code and found that you create Quote data in hard-coded way, not fetching from spreadsheet. Is there any reason for this?

For me, hardcoding is the best way too keep that data static and to protect it from accidental changes.  I prefer that over reading from files, but can see using CSV or other delimited files too.

I really only use the Excel file as a basis for doing manual calculations.

LeeDongGeon1996:
Finally, it works on Azure Pipelines. Before that, can you check other PRs that makes this changes work correctly?

> I suspect reading from Excel might not work in environments where there is no native MS Office tools.

Yeah, MS Office tools are usually hard to be used on other systems, not Windows. But `.xlsx` extension is just one of file extensions, I'm sure that there is always programmatic way to read file regardless of which OS system we are using. So I think we don't need to convert `History.xlsx` to other delimited text file. Of course, totally agree with that C# test code should read data directly from the file.

DaveSkender:
I'll review this more later, but a few quick items:

- [ ] Update test results category names (as shown in build results)
- [ ] Update test names to be more specific
- [ ] Save Wheel package that will eventually be uploaded to PyPI
- [ ] Evaluate code coverage results in build results, to see if there's a way to show them separately for Python

LeeDongGeon1996:
Could you give me more details about the first two things?

> *Evaluate code coverage results in build results, to see if there's a way to show them separately for Python*

Seems like we need to merge two separated coverage reports in one.  ==> [Publish Code Coverage Results task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-code-coverage-results?view=azure-devops#is-code-coverage-data-merged-when-multiple-files-are-provided-as-input-to-the-task-or-multiple-tasks-are-used-in-the-pipeline)

DaveSkender:
For the first one, now that we have Python test results for two versions and the existing .NET test results, these batch names should be more descriptive.

image

Suggest:
- Python 3.8 Tests
- Python 3.9 Tests
- .NET Tests

For the second one, the actual names of tests might be better.  But to be fair, I have the same problem with existing C# tests.  Not too worried about this one.

image

Regarding the code coverage, I'd not try to mix them up between .NET and Python tests.  The .NET code coverage is most important.  Don't try to publish the Python code coverage data if it can't be done as a separate concern.  As long as the build log shows Python coverage, that's good enough.

LeeDongGeon1996:
Do you think we should run tests also in Windows environment as well as Linux system?

DaveSkender:
> Do you think we should run tests also in Windows environment as well as Linux system? (For Python wrapper)

Probably not, but I'm not sure.  These are functional tests, so we checked the code, not OS compatibility.  My assumption is that a final build package will have a universal OS compatibility.

LeeDongGeon1996:
Yes, but as we had been suffering from selecting .NET version to be used for `pythonnet`, I think it is worth to check that `pythonnet` works correctly with .NET version we choose on both system.

LeeDongGeon1996:
[[`436f749`]()] Are you able to update `setup.py` and create PR to this branch(`LeeDongGeon1996:test/Azure-Pipelines`)?
The `setup.py` is the file containing the information for production package and shown on PyPi website. You can refer this [test PyPi](https://test.pypi.org/project/Here-is-package-name/0.0.1/) page to figure out which value is used for what.

DaveSkender:
Yes, I'll take a look tomorrow at `setup.py` and get that extra info for you.

DaveSkender:
@LeeDongGeon1996, it doesn't look like GitHub wants me to create feature branches on your Fork to do a PR like that, so I just pushed here.  Let me know if you need anything else for setup.

LeeDongGeon1996:
>  I just pushed here.

This is better. Thanks!! 
Could you check #438? The dll has to be in _cslib to be included in build package.

DaveSkender:
> > I just pushed here.
> 
> This is better. Thanks!!
> Could you check #438? The dll has to be in _cslib to be included in build package.

Just merge #438 into this branch instead of having it as a separate PR to master; it's a related change.

LeeDongGeon1996:
Done! Please check the whole process. The build artifact are saved in `packages`.

DaveSkender:
image

- Can we avoid permanently saving this DLL folder?  I suspect we don't need it forever.
- Are these Python `.tz` and `.whl` files supposed to use the same project name (e.g. - vs _)?
- What is `-none-any` meaning?

LeeDongGeon1996:
I'm kind of busy this week. I'll work on it this weekend.

LeeDongGeon1996:
> Can we avoid permanently saving this DLL folder? I suspect we don't need it forever.

(Working on it)

 > Are these Python .tz and .whl files supposed to use the same project name (e.g. - vs _)?

According to [PEP 427](https://www.python.org/dev/peps/pep-0427/#file-name-convention),
the wheel file name has its naming convention. It is `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl` format, which sections are divided with `-`. In our case, the {distribution} has long name and is consisted with several words. So I don't think it's not good idea to change its default name.

 >  What is -none-any meaning?

According to [PEP 427](https://www.python.org/dev/peps/pep-0427/#file-name-convention),
 * none — is not OS-specific
 * any — suitable to run on any processor architecture
 

>  Can the PythonBuild part start after DotNet, in parallel with Python tests? Might go faster.

(Working on it)

LeeDongGeon1996:
1. There is a [warning](https://dev.azure.com/skender/Stock.Indicators/_build/results?buildId=2577&view=logs&j=14f058a8-c161-5eca-312a-911a13e3bdba&t=90bfc812-3f8b-5f00-c946-84b96592c9d6&l=41) while building Python Wheel: if 'maintainer' supplied, 'maintainer_email' must be supplied too.
2. Why don't you specified `author` field in `setup.py`?

DaveSkender:
> * There is a warning while building Python Wheel: if 'maintainer' supplied, 'maintainer_email' must be supplied too.
> * Why don't you specify `author` field in `setup.py`?

I was hoping to avoid exposing my personal email address.  I used Maintainer instead of Author because there are many contributing authors for open source projects.  Specs say you can use either.  I might be overthinking this :smile:

LeeDongGeon1996:
> was hoping to avoid exposing my personal email address. I used Maintainer instead of Author because there are many contributing authors for open source projects. Specs say you can use either. I might be overthinking this 😄

So kind of you. Maybe it might be better to place your name in `author` field and other main contributors in `maintainer` fields. But it's entirely up to you.

By the way, Is there anything to do more about this?

.github/build.main.yml
```diff
@@ -4,95 +4,238 @@ trigger:
...
+        inputs:
+          testResultsFiles: '$(PythonWrapperBaseDirectory)/tests/test-results.xml'
+          testRunTitle: 'Python $(python.version) Tests'
+
+    - job: Build
+      variables:
+        PythonWrapperBaseDirectory: "$(Build.SourcesDirectory)/wraps/python"
+        ImageName: "ubuntu-latest"
+        DotNetVersion: "netstandard2.0"
+        PythonVersion: "3.8"
```
DaveSkender, line 184: When we're building this package for PyPI, does it matter that we're only using one version here?  Or is there an option or need to also build for v3.9?  I'm unclear on how PyPI works in this regard.

.github/workflows/codeql-analysis.yml
```diff
@@ -30,9 +29,23 @@ jobs:
...
+          python-version: "3.x"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          if [ -f requirements.txt ];
+          then pip install -r requirements.txt;
+          fi
+          # Set the `CODEQL-PYTHON` environment variable to the Python executable
+          # that includes the dependencies
+          echo "CODEQL_PYTHON=$(which python)" >> $GITHUB_ENV
```
DaveSkender, line 50: Can you check to see if I did this right, for CodeQL?  Just trying to add Python here to security scanning.

wraps/python/setup.py
```diff
@@ -0,0 +1,35 @@
+import setuptools
+
+with open("README.md", "r", encoding="utf-8") as fh:
+    long_description = fh.read()
+
+setuptools.setup(
+    name="skender-stock-indicators",
+    version="0.0.1",
+    author="Dave Skender",
```
DaveSkender, line 9: I've added you here as maintainer [for Python wrapper] if you're okay with that role designation.  From what I can tell from the specs, email is not required; even if it's producing an informational warning.  I've left it out because I'd rather not dox myself - there are other methods to contact us in docs.

DaveSkender (changes requested):
Looks good overall.  I've only made some minor tweaks to the build to ensure Python only starts after the DotNet build (since it relies on the DLL).  Python Tests and Build occur in parallel as they're not co-dependent.  I've left a few comments for you to review as well.

.github/workflows/codeql-analysis.yml
DaveSkender: It looks like it's running okay, but wanted a second opinion.

.github/build.main.yml
LeeDongGeon1996: I think the version of Python doesn't matter when we build. But to make it be sure, I posted about it on Stack Overflow. 
 * https://stackoverflow.com/questions/67761075/which-python-version-to-be-used-for-building-pypi-package

LeeDongGeon1996: I found a problem about minimum Python version of our package. It seems like that `pythonnet` supports Python 3.5 and higher. But when I tested on Azure Pipelines, errors occurred when Python 3.5~3.7. I just thought that versions of Python are not supported, but it was wrong. So we need to figure out. I think it's better to pass it for now and deal with this problem on new Github issue because it looks like I need more time to test.

.github/workflows/codeql-analysis.yml
LeeDongGeon1996: It looks like there is ostensibly no problem. But, I think it's not working as you originally intended. You are installing Python dependencies manually. 
The problems are:
1. We don't have `requirements.txt`.(But, I think it's better to make this file.)
2. Needs to set `setup-python-dependencies: false`

You can check official docs here: [Analyzing Python dependencies](https://docs.github.com/en/code-security/secure-coding/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#analyzing-python-dependencies).

wraps/python/setup.py
LeeDongGeon1996: > there are other methods to contact us in docs.

Yep, you are right. 

> I've added you here as maintainer [for Python wrapper] if you're okay with that role designation. 

Thanks, I love to be there. But could you fix my name to `Dong-Geon Lee`. My name looks not good for readability to native English speaker in my experience. So I usually write my name with `-`.😂

.github/build.main.yml
DaveSkender: Yah, go ahead and create an issue to track with mention of this conversation.  We have some time before we're publishing, but will need to figure it out along the way.  No need to hold this PR up.

.github/workflows/codeql-analysis.yml
DaveSkender: Yah, I copied code from there but missed the 2nd point.  I've also removed the `requirements.txt` option since we're not using it.  I think this is working now.

DaveSkender (dismissed):
Anything else you want to take care of before we merge?

LeeDongGeon1996:
I've just made some minor changes. Please check the commits below, and merge if there is nothing more to discuss.

 *  [[`a6317b4`](https://github.com/DaveSkender/Stock.Indicators/pull/424/commits/a6317b47c083533b2ac7964787755bd4f9b381d0)] Add dependencies in `setup.py`
  *  [[`c1eb59c`](https://github.com/DaveSkender/Stock.Indicators/pull/424/commits/c1eb59cb47f383a0ba373f24c35350fb23725784)] Add `requirements.txt`
   *  [[`0332fb1`](https://github.com/DaveSkender/Stock.Indicators/pull/424/commits/0332fb12d37d853d17654daeba76b73475af054d)] Update `codeql-analysis.yml` to install dependencies.

DaveSkender (approved):

Issue #530, opened 2021-08-17
Aggregate using TimeSpan


heavymanto:
Hi, I have a timespan value and i would want to convert him in Periodsize, can i implent this function?
Thanks you

DaveSkender:
👍🏼  Seems simple enough to add this to our library of related helper methods.

heavymanto:
Hi, i have implemented funtion and test method. Can a push? In what branch?
thanks you

DaveSkender:
Create a PR from the feature branch in your Fork to the `main` branch in this repo and I’ll take a look.

DaveSkender:
@heavymanto is the use case for this mostly about using a known `TimeSpan` in the `quotes.Aggregate(PeriodSize periodSize)` function?  I ask, because it would be cleaner to simply provide an override like `quotes.Aggregate(TimeSpan timeSpan)` instead.

heavymanto:
hi Dave, yes this is my use case. The use of the overranide is an excellent choice. You want me to change the implementation by overriding quotes.Aggregate (TimeSpan timeSpan)?

DaveSkender:
Yah, if that's your use case, I think it makes sense to do that instead ... it will make it one less step for users.  Thinking about this more, I was thinking that this `.ToPeriodSize()` would have really only one use, and it's probably better to just allow a different parameter type here instead.

heavymanto:
What do you think if in the override Aggregate(TimeSpan timeSpan) call the extension method? In this way with only one method it can be used in multiple scenarios?

`funtion Aggregate(TimeSpan timeSpan)
  return Aggregate(timeSpan.ToPeriodSize())`

DaveSkender:
That can work if `timeSpan` is a compatible value for `periodSize`.  I'm generally okay with skipping the intermediate conversion to `PeriodSize` for a few reasons: 1) `Aggregate` natively uses `TimeSpan` internally, 2) it would be a double conversion back to itself, and 3) using `timeSpan` without forcing `periodSize` would allow users to customize -- for example, if they wanted 45 second periods.

heavymanto:
Ok, thanks for the clarification. I study code, and them propose a solution

heavymanto:
ok, I have dev quotes.Aggregate(TimeSpan timeSpan) funtion and open PR.
Thanks

DaveSkender:
This was release in version 1.17.1

PR #544, opened 2021-09-03
add ToPeriodSize

heavymanto:
## Description

Implemented extension function 

`internal static TimeSpan ToTimeSpan(this PeriodSize periodSize)`

I had to use a multi statement if as the `switch` statement does not support the TimeSpan type

P.S. Is my first contributions and i always i use TFS. I hope it is to your liking.

Implements #530 

## Checklist

- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings and running code analysis does not produce any issues
- [ ] New and existing unit tests pass locally and in the build (below) with my changes

## Acknowledgements

- [ ] I have read and understand [the Apache 2.0 license](https://opensource.org/licenses/Apache-2.0)
- [ ] I agree to the [Developer Certificate of Origin (DCO) 1.1](https://developercertificate.org)

### DIFF
--- a/src/_Common/Functions.cs
+++ b/src/_Common/Functions.cs
@@ -105,6 +105,36 @@ namespace Skender.Stock.Indicators
             };
         }
 
+        internal static PeriodSize? ToPeriodSize(this TimeSpan timeSpan)
+        {
+            PeriodSize? _periodSize = null;
+
+            if (timeSpan == TimeSpan.FromMinutes(1))
+                _periodSize = PeriodSize.OneMinute;
+            if (timeSpan == TimeSpan.FromMinutes(2))
+                _periodSize = PeriodSize.TwoMinutes;
+            if (timeSpan == TimeSpan.FromMinutes(3))
+                _periodSize = PeriodSize.ThreeMinutes;
+            if (timeSpan == TimeSpan.FromMinutes(5))
+                _periodSize = PeriodSize.FiveMinutes;
+            if (timeSpan == TimeSpan.FromMinutes(15))
+                _periodSize = PeriodSize.FifteenMinutes;
+            if (timeSpan == TimeSpan.FromMinutes(30))
+                _periodSize = PeriodSize.ThirtyMinutes;
+            if (timeSpan == TimeSpan.FromHours(1))
+                _periodSize = PeriodSize.OneHour;
+            if (timeSpan == TimeSpan.FromHours(2))
+                _periodSize = PeriodSize.TwoHours;
+            if (timeSpan == TimeSpan.FromHours(4))
+                _periodSize = PeriodSize.FourHours;
+            if (timeSpan == TimeSpan.FromDays(1))
+                _periodSize = PeriodSize.Day;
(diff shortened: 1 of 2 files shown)


src/_Common/Functions.cs
```diff
@@ -105,6 +105,36 @@ internal static TimeSpan ToTimeSpan(this PeriodSize periodSize)
             };
         }
 
+        internal static PeriodSize? ToPeriodSize(this TimeSpan timeSpan)
```
DaveSkender, line 108: `internal` will not be accessible by library users, so you'll have to make this `public`

src/_Common/Functions.cs
```diff
@@ -105,6 +105,36 @@ internal static TimeSpan ToTimeSpan(this PeriodSize periodSize)
...
+        {
+            PeriodSize? _periodSize = null;
```
DaveSkender, line 110: I've not been using underscores for private variables in our [naming convention](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines); suggest removing for consistency.

DaveSkender (changes requested):
Generally looks okay.  A couple initial minor things below.  I may also update the guide docs to convey this (unless you want to do it).

Can you put this into a feature branch instead of your `main`, something like `period-size`?  That will make it easier for me to contribute to your work.  It's not normal to do the PR from your `main` branch, though it works okay as long as you're not commingling.  If you change the branch name, you'll have to edit this PR to source from the new feature branch.

src/_Common/Functions.cs
```diff
@@ -105,6 +105,36 @@ internal static TimeSpan ToTimeSpan(this PeriodSize periodSize)
...
+
+            if (timeSpan == TimeSpan.FromMinutes(1))
```
DaveSkender, line 112: If this doesn't support `switch`, it might be more efficient to just delete `_periodSize` and just do:

```csharp
if (timeSpan == TimeSpan.FromMinutes(1))
  return PeriodSize.OneMinute;
...
```
This will avoid subsequent evaluations.

src/_Common/Functions.cs
heavymanto: can also public ToTimeSpan and RoundDown method?

src/_Common/Functions.cs
heavymanto: I started developing as you suggested, however there is always a need for a default return otherwise it won't compile. Otherwise I could use a null retrun, which only occurs if the precedents are not satisfied, can it go?

`if (timeSpan == TimeSpan.FromMinutes(1))
return PeriodSize.OneMinute;
...
return null`

DaveSkender: Yes, adding a `return null;` at the end would satisfy that need.

src/_Common/Functions.cs
DaveSkender: I'm not a fan of exposing all these other internals if we can avoid it.  If it's necessary, go ahead and make them `public`.

DaveSkender:
Replace by #545

PR #618, opened 2021-11-13
fix: ZigZag when no thresholds met

### LINKED ISSUE #616: ZigZag should not plot line when no thresholds are met
## the [potential] problem

- [x] Confirm that the last line of ZigZag is not plotting extra interim points before thresholds are met.
- [x] Add extra unit tests that confirm unusual scenarios - see [eval-zigzag branch](https://github.com/DaveSkender/Stock.Indicators/tree/eval-zigzag).

This is a follow-on to #600 

## to reproduce

See results from the unit test in the above branch.  There are some extra zigs at the end, but it could be my unit test is bad or am not interpreting it correctly.

## expected behavior

The last line is straight to the last point if it is not long enough to reach a threshold.

DaveSkender:
## Description

Exploratory investigation to see if the last line is behaving correctly given a new aberrant dataset.
Fixes #616

## Checklist

- [x] New and existing unit tests pass locally and in the build (below) with my changes


DaveSkender:
Using the data provided in the linked issue, there is never a full 5% reversal; therefore, no long should have drawn at all!
I'm fixing that here.  What was happening is that it got to the end of the history range and simply plotted the incomplete interim results, which is why you see some odd danglers at the end.

image

DaveSkender:
The analysis is basically doing this: 1) find the lowest extreme point, then 2) try to find the 5% reversal.

image

Given this data, it never increases by 5% and that first low point is unconfirmed; therefore, no line should show.

PR #681, opened 2022-01-17
opt-out of `BadQuotesExceptions`

### LINKED ISSUE #656: Damn the [convergence] torpedos
As a user who doesn’t care about convergence-related precision errors (see #215), I’d like to opt-out of the related extended minimum quote history threshold.

Requested from @Jammooly1 in [RSI discussion](https://github.com/DaveSkender/Stock.Indicators/discussions/224#discussioncomment-1784202)

This might be a global setting or a per-usage setting, but probably not an indicator parameter.  Consider also returning a null set series for incalculably small history.

Investigate:

- [x] [class attributes](https://docs.microsoft.com/en-us/dotnet/standard/attributes/writing-custom-attributes).  Conclusion: too complicated to both implement and use.
- [x] [IDisposable](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement) "using" for one-off usage.  Conclusion: only works with limited use cases and is not compatible with static classes.
- [x] other methods for changing runtime settings without the use of a config file.  Conclusion: trying with static global configurations, but don't love it due to thread-safety concerns; also, cannot find good alternatives and we really can't use config files as this somewhat relies on the caller application.

An example use case is to allow a new market issue (IPOs) to have indicators as early as they can be calculated, even if they're not entirely converged yet.

image

Open concerns:

- [ ] Running into constraints since we're using static classes that do not support independently threaded configurations, so the only option is to somehow gracefully set global configuration parameters.

DaveSkender:
### Description

Adds a general convention to allow user settings that can override default behaviors of the library.  This specific change allows users to GLOBALLY opt-out of using `BadQuotesException`, so when the quantity of quotes provided is below the warmup and convergence period thresholds, it will return results without throwing an exception.

This is a PREVIEW feature.  See #656 for more information.

```csharp
// STRONGLY recommend you do this in your startup procedures so it does not
// cause a race condition; you should not use this on a per-call basis as it is GLOBAL impact!

// turn off the exception (global setting)
Indicator.UseConfig(
    new IndicatorConfig
    {
        UseBadQuotesException = false
    });

...

// fetching [not-enough] quotes (3 bars)
var quotes = TestData.GetDefault(3);

// returns 3 dated records where Ema = null
// normally this would throw a BadQuotesException
var results = quotes.GetEma(5);
```

### To do

- [x] Consider abandoning due to the lack of thread-safety.  This might work better as an appsettings.json configuration but is quite volatile when called inline like it is shown above.  I don't want to use config files.

### Checklist

- [x] I have added or updated unit tests that prove my fix is effective or that my feature works, and achieves sufficient code coverage.  New and existing unit tests pass locally and in the build (below) with my changes


DaveSkender:
Abandoning this approach due to the quirkiness of how this can cause problems for users.

PR #707, opened 2022-02-04
update

BlankaKorvo:


DaveSkender:
@BlankaKorvo, just peeking at your open-source projects, you might benefit from just using our NuGet package directly, instead of a ported project reference.  Our compiled NuGet package is probably 2 times faster than compile on demand, especially when working in Debug mode.

PR #741, opened 2022-03-11
add back legacy .NET framework 4.6.1

DaveSkender:
Still seeing people using older frameworks, so adding back support for .NET 4.6.1 targeting.

- [x] re-consider before merging.  if this is not supported in newer IDEs, maybe it’s not a good idea


DaveSkender:
This isn’t going to work with C# 10 and newer build tool.  Time to move on and upgrade if you’re still using the old .NET Framework.  I tried, but it’s not worth the hassle.

PR #752, opened 2022-03-20
chore: code cleanup

DaveSkender:
Resolving some minor code analysis findings from LGTM.  Nothing noteworthy.


DaveSkender:
LGTM for C# is still producing a few false positives and interestingly the "new" alert was generated without any of the related code being modified.

PR #798, opened 2022-05-07
add OHLC4

tiger2014:
add OHLC4 in Quote

### DIFF
--- a/src/_common/Quotes/Quote.Aggregates.cs
+++ b/src/_common/Quotes/Quote.Aggregates.cs
@@ -33,6 +33,7 @@ public static partial class HistoricalQuotes
                 Low = x.Min(t => t.Low),
                 Close = x.Last().Close,
                 Volume = x.Sum(t => t.Volume)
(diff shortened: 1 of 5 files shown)


DaveSkender:
I have some guidance on how you can [create your own custom quote class](https://daveskender.github.io/Stock.Indicators/guide/#using-custom-quote-classes).
I have no use for this modification in [the package](https://www.nuget.org/packages/Skender.Stock.Indicators) itself.

DaveSkender:
By the way, I'm probably reverting the nullable Quote mods because they have an adverse impact on performance.  Those nullable Quote changes were never released.  I'd recommend using the [released library package](https://www.nuget.org/packages/Skender.Stock.Indicators) instead of the raw source code; unless you're contributing to the library, of course.

DaveSkender:
If your goal is to simply have and use OHLC4, just wait a bit for me to catch up.  I'm working towards some new features that will allow users to do something like this:

```csharp
quotes
  .Use(CandlePart.Close)
  .GetRsi(14);
```

tiger2014:
Thank you, DaveSkender.

PR #802, opened 2022-05-09
indicator chaining

### LINKED ISSUE #630: Get the Slope of any indicator in the library
For example if requesting any indicator the output is typically a series of time dependent values. Is there a "easy" solution that I can implement to calculate the first derivative (slope) of that indicator output or is that something that has to be manually coded for each indicator where I need to get the IEnumerable result then calculate each slope.

Would it be possible to add a shared method that can be inherited and can be called on any indicator output that will return a list of slope values at each possible point? For example if called on SmaResult output it will return a list of n-1 values of slope for each point except the first point of course.

DaveSkender:
### Description

Resolve #630 and adding general indicator chaining so you can do things like:

```csharp
# choose any candle part or variant
# to override the default
var results = quotes
  .Use(CandlePart.OHLC4)
  .GetRsi(14);
```
```csharp
# slope of any indicator
var results = quotes
  .GetMacd()
  .GetSlope();
```
Note: not all combinations are available as a secondary combination since they require the full quote for calculation (e.g. ATR, Heikin-Ashi require multiple candle parts).

### Breaking changes

- numerous result values and some parameters changed from `decimal` to `double`
- result class is now `sealed` and cannot be used as a derived class (required for reusable chaining), though you can just replicate the class properties
- IMPORTANT!:  Beta and Price Relative Strength (PRS) have swapped positions of Eval and Base (Market) quote parameters positions, for consistency.  Since these have the same data type profile, it does not naturally produce a warning!  A warning is forced on v2.0.0 only as a precaution but will be removed in the next released version.
- `GetSmaExtended()` was renamed to `GetSmaAnalysis()`.  We'll expand on and improve extended analysis broadly in future versions.  See #320 on how this may change in the future.


LeeDongGeon1996:
I've thought about the way to add these features(chaining, streaming) and figured out that it needs substantial refactoring.
> It might be better to get the wrapper to a stable 1.0 before starting on some of this.

I think so. And if possible, I'd like to version up pythonnet so that the library can support .NET 6.0. :)

PR #824, opened 2022-06-19
streaming preview, prototype

DaveSkender:
### Description

A preview of streaming use cases for quotes, `.Use`, EMA, and SMA only, for prototyping all of our scenarios.

This is the first in a series of preview releases that implement streaming use cases.  We expect possibly 5-10 preview pre-release versions before we make an official v3 package.  They are all experimental and will be quite volatile with breaking changes and conventions.

```csharp
# initialize quotes provider,
# can be extended to integrate with your
# quote stream or WebSocket source
QuoteProvider provider = new();

# subscribe individual indicators as observers
EmaObserver ema = provider.GetEma(20);
SmaObserver sma = provider.GetSma(13);
[..]

# these results are automatically updated
IEnumerable<EmaResult> emaResults = ema.Results;
IEnumerable<SmaResult> smaResults = sma.Results;

# when quotes are added in a separate process
provider.Add(quote);

# and also works with `.Use` chain
EmaObserver emaAlt = provider
    .Use(CandlePart.HL2)
    .GetEma(20);

# stop subscriber auto-updates 
# by unsubscribing each observer
ema.Unsubscribe();

// stop providing quotes to everyone
// will unsubscribe all gracefully
provider.EndTransmission();

# feedback appreciated
```

### Use cases supported (can do)

- transmit streaming quotes as a provider to indicators
- `.Use`, EMA, and SMA indicators only
- see WebSocket live test case: `/tests/observe/`

### Known limitations (cannot do)

- we do not handle late, out-of-order arrivals of quotes from streaming feeds, they are ignored
- indicators of indicators not supported, but we do support `.Use` chains
- incomplete documentation
- :warning: if you hookup to a **trades** stream, you may get many different quotes for an intended period, yet only an aggregate of them represents a bar (they are individual trades, not candle updates).  Use caution when selecting a provider.

### What we'll do next

- add and simplify indicator of indicator chaining
- update documentation
- troubleshoot out-of-sequence late quotes and chain arrivals


codebeaulieu:
This would be super handy for use with web sockets. Reactive Extensions internals would be a major plus... possibly consider [Dynamic Data](https://dynamic-data.org) as well, which is what I'm currently using.

DaveSkender:
> This would be super handy for use with web sockets. Reactive Extensions internals would be a major plus... possibly consider [Dynamic Data](https://dynamic-data.org) as well, which is what I'm currently using.

For sure.  I'm thinking of WebSocket specifically for this one.

Thanks for the reference on Dynamic Data -- it might be easier than simply using the native `IObservable` alone, something I will explore further.  I'll have to weigh my appetite for third-party packages; I like the idea of continuing without any.

DaveSkender:
I've come back to this and am currently working through the complexities ... like dealing with out-of-sequence late quote arrivals through chains of dependencies 😮.  @codebeaulieu and @danbopes, I'd love some early feedback on the public interface UX depicted in the description above if you have any.

martonb:
Just my humble opinion: you started to solve too many issues at once?
Personally, i would just have started with creating indicators that don't need to recalculate the full sequence every time a new quote comes in. Only later on create the extra code needed to connect to web sockets, or leave people to make their own hook-up connections, maybe by implementing an interface provided by your API?

Then, for my use case, I would be able to:
- request historic quotes from an arbitrary exchange, using whatever method (eg: REST request), from this moment, backwards N + warmup periods count quotes
- create the indicator, initialized with these quotes
- now I already have a valid indicator value calculated
- when a new quote notification is received from the exchange (via pure TCP/IP (protobuff) or WebSocket (JSON/XML), etc) your API user can process that and pass it to the indicator
- the indicator might delete the oldest quote from the sequence, then progressively calculate the new value
- the user can use the new calculated indicator value

DaveSkender:
> Just my humble opinion: you started to solve too many issues at once? Personally, I would just have started with creating indicators that don't need to recalculate the full sequence every time a new quote comes in.

This is some good general advice.  Thank you.  My goal here is to simply do a single tracer the whole way through before widely implementing across indicators.  Once I get the basic concept working well, I'll think about how to apply some useful increments in smaller forms.

DaveSkender:
> I assume they are internal, as not production ready?

That’s right.  It’s only in the feature branch right now.  I’ll release a preview package soon, for people to try it out.

PR #827, opened 2022-06-19
refactor: internal indices

### LINKED ISSUE #826: zero doesn't equal one here
Why not set `i=1` then you only have to subtract once instead of add twice

https://github.com/DaveSkender/Stock.Indicators/blob/4589861fe8187b90d10c0460c0d4b7f89450d30e/src/a-d/BollingerBands/BollingerBands.Series.cs#L18

https://github.com/DaveSkender/Stock.Indicators/blob/4589861fe8187b90d10c0460c0d4b7f89450d30e/src/a-d/BollingerBands/BollingerBands.Series.cs#L27

https://github.com/DaveSkender/Stock.Indicators/blob/4589861fe8187b90d10c0460c0d4b7f89450d30e/src/a-d/BollingerBands/BollingerBands.Series.cs#L33

unless I'm missing something, Yes, I know we usually start at 0.

DaveSkender:
### Description

Resolves #826
and generally improving the readability of internal use of indices.  This is a follow-on to the prior replacement of the unneeded and disliked `index` helper variable.

### To do

- [ ] apply convention consistently to all indicators

### Checklist

- [ ] My code follows the existing style, code structure, and naming taxonomy
- [ ] I have performed a self-review of my own code and included any verifying manual calculations
- [ ] I have added or updated unit tests that prove my fix is effective or that my feature works, and achieves sufficient code coverage.  New and existing unit tests pass locally and in the build (below) with my changes
- [ ] My changes generate no new warnings and running code analysis does not produce any issues
- [ ] I have made corresponding changes to the documentation

### DIFF
--- a/src/a-d/BollingerBands/BollingerBands.Series.cs
+++ b/src/a-d/BollingerBands/BollingerBands.Series.cs
@@ -12,25 +12,26 @@ public static partial class Indicator
         ValidateBollingerBands(lookbackPeriods, standardDeviations);
 
         // initialize
-        List<BollingerBandsResult> results = new(tpList.Count);
+        int length = tpList.Count;
+        List<BollingerBandsResult> results = new(length);
 
         // roll through quotes
-        for (int i = 0; i < tpList.Count; i++)
+        for (int i = 1; i <= tpList.Count; i++)
         {
-            (DateTime date, double value) = tpList[i];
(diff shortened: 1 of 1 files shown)


DaveSkender:
I think I’ll refactor these opportunistically instead of doing it as a separate change.  There’s no significant cost of delay.

PR #847, opened 2022-07-16
Modify GetRsi() to utilize ienumerable/improve performance

danbopes:
This is my first attempt at improving optimization of the `GetRsi()` function, and offering guidance for best practices when enumerating values.

This pull request is mainly a springboard to hopefully rewrite certain parts in more meaningful ways to avoid full iterations to lists to calculate various indicators. This pull request will have little improvement mainly because of:

```
public static IEnumerable<RsiResult> GetRsi<TQuote>(
        this IEnumerable<TQuote> quotes,
        int lookbackPeriods = 14)
        where TQuote : IQuote => quotes
            .ToBasicTuple(CandlePart.Close)
            .CalcRsi(lookbackPeriods);
```

`ToBasicTuple()` returns a list, which forces a full iteration of all the items in the `IEnumerable`, which can be *highly* inefficient when dealing with large collection sets. In the case of `GetRsi()`, a full enumeration is not required.

For example: If I wanted to simply calculate the latest RSI value in a given quote set, this function will take 1000 data points (Which I could be streaming from a file), put them all into memory, iterate through all of them, returns a list of results, and then grab the last item in the list.

If `.ToBasicTuple()` simply returned an `IEnumerable`, then we could start streaming the data one item at a time to `CalcRsi()`, it could calculate the RSI, and discard all but the last value. The entire memory consumption would be essentially at most 2 `RSIResult`'s (One stored in the loop as the last value, and one in the loop checking to see if it's the last value, and overwriting it).


DaveSkender:
It will be interesting to see what you do here after you can get the unit tests to pass (calculate the same values as before modification).

We’re instrumented to do performance tests (see [docs](https://github.com/DaveSkender/Stock.Indicators/blob/main/docs/contributing.md#performance-benchmarking)), so we’ll be able to see how much this improves. 

I’ve seen performance improvement in my current work on incremental designs for streaming use cases, but would like to see where your approach goes to see if it’s better.

DaveSkender:
I think some of the stuff in your other branch looks promising.  I suspect the minor performance improvement from 55us to 50us here is primarily due to how you're iterating on the tuple slightly differently -- a potential improvement.  Most of the other changes _appear_ to be just minor rearrangements that probably don't help much.

Even if in a different PR, see if you can get RSI to use the same public API, pass the unit tests -- to indicate you're still doing it right, and don't worry about preserving the old internals.

DaveSkender:
And just an FYI, here are some helpful tips when working in a fork for the purpose of contributing to an upstream repo: #503 
You've polluted your `main` branch, so it'll be harder to pull updates from the source or to create new fresh feature branches.

DaveSkender:
In the end, I think what you're recommending the use of incremental calculations over full time-series calculations.  This is something I'm actively working on in the `streaming-preview` branch and PR #824.  I'm also seeing notably better performance when incorporating it back into the time-series models.

```text
for EMA
|       Method |     Mean |    Error |   StdDev |
|       GetEma | 57.81 us | 0.896 us | 0.839 us |
| GetEmaStream | 10.37 us | 0.112 us | 0.105 us |
```

DaveSkender:
@danbopes let me know if you're still working on this.  Closing due to staleness, but can reopen.

PR #937, opened 2022-11-02
chore: update docs

DaveSkender:
Documentation updates only.  No code changes.


LeeDongGeon1996:
If you don't mind, could you apply these changes on docs to Python one??

DaveSkender:
> If you don't mind, could you apply these changes on docs to Python one??

Sure.  What’s your impression of the recent indicator page updates?  I have mixed feelings on it.

BE5138B3-E1FC-4BB0-B586-4977F1C2CC5B

LeeDongGeon1996:
Thanks. All I'd like to ask you to care is just sync with .NET docs.

> What’s your impression of the recent indicator page updates?

I also have mixed feelings about it, but I think that the previous one is better only for listing indicators. And current change is better for showing indicator categories.

PR #970, opened 2022-12-16
code analysis fixes

DaveSkender:
### Description of changes

Some new code analysis updates have highlighted a few areas for improvement with `public` APIs for safety and extensibility.

### Changes
- [x] Treat warnings as errors in the build
- [x] Add missing validation for `Numerix` classes that are now public (`Slope` and `StdDev`)
- [x] Convert public return values for quote classes from `List` to generic `Collection` types
- [x] Convert public return values for utility classes from `List` to generic `Collection` types

### :warning: **breaking changes** :warning:

- class `Numerics` was renamed to `Numerix` to avoid conflict with `System.Numerics`

These now return `Collection` instead of `List` and were renamed:

- renamed `results.ToSortedList()`  »  `results.ToSortedCollection()`
- renamed `quotes.ToSortedList()`  »  `quotes.ToSortedCollection()`
- renamed `tuple.ToSortedList()`  »  `tuple.ToSortedCollection()`
- renamed `.ToTuple()` » `ToTupleCollection()`


DaveSkender:
I think I over-engineered the use of `Collection` to replace `List`.
This results in an average 17% reduction in indicator performance with only a few small improvements.
No bueno.  Back to the drawing board!

PR #983, opened 2022-12-24
add Beta, Corr, Prs chaining options

DaveSkender:
### Description

> Is there a reason there's no API of `IResusableResult` for Beta? Since the Python wrapper will not use tuple API, so it needs a chaining method using the `IReusableResult` interface.  -- @LeeDongGeon1996

The only reason is that it gets complicated and may cause some confusion for users trying to map this correction.  Beta currently requires same-length quotes and chained items often come in with different lengths with potentially mismatched dates.  I think we can use the internal `.SyncIndex()` methods to resolve some of these challenges, potentially.

### To do

- [x] add new API use cases for Beta, Correlation, and Prs
- [x] update docs for `Beta`, `Corr`, `Prs`, etc to clarify the limitation

### Checklist

- [x] I have put comments in my code, particularly for hard-to-understand areas
- [x] I have performed a self-review of my code and included any verifying manual calculations
- [x] I have added or updated unit tests that prove my fix is effective or that my feature works and achieves sufficient code coverage.  New and existing unit tests pass locally and in the build (below) with my changes
- [x] My changes generate no new warnings or other code analysis issues


DaveSkender:
This was implemented in v2.4.8

PR #1021, opened 2023-02-11
add 'Visualize BackTest Result'

tiger2014:
### Description

Let the backtest result be visualized with simple method


tiger2014:
This is very a simple sample. Hope someone knowing more about the frontend can add a pane for indicators (i.e. MACD, RSI) on the chart. Thanks!

DaveSkender:
The only place I’d consid

DaveSkender:
The only place I’d consider putting something like this is in the example project.  I’m not keen on putting front-end rendering pieces here.  It might make a good, but separate, add-on package as part of #484

DaveSkender (changes requested):
This can’t be part of the core library since it’s out of scope.  See notes above.

tiger2014:
I see

DaveSkender:
Another option here is to just make a standalone example and upload it to your original "show discussion" here: #1005

DaveSkender:
Transferred to discussion

PR #1046, opened 2023-04-17
Remove default value for StarcBands

Temppus:
### Description

Remove default value for **_smaPeriods_** argument for StarcBands indicator. Previous 20 was too high and there is no general default value so force users to provide some.

See https://github.com/DaveSkender/Stock.Indicators/discussions/1045

### DIFF
--- a/tests/indicators/s-z/StarcBands/StarcBands.Tests.cs
+++ b/tests/indicators/s-z/StarcBands/StarcBands.Tests.cs
@@ -66,13 +66,13 @@ public class StarcBandsTests : TestBase
     public void NoQuotes()
     {
         List<StarcBandsResult> r0 = noquotes
-            .GetStarcBands()
+            .GetStarcBands(10)
             .ToList();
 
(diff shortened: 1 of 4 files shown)


tests/indicators/s-z/StarcBands/StarcBands.Tests.cs
```diff
@@ -66,13 +66,13 @@ public void BadData()
     public void NoQuotes()
     {
         List<StarcBandsResult> r0 = noquotes
```
DaveSkender, line 68: To get these tests to pass, you'll have to put them in as 20 here, which is okay for the tests since any number will do.

DaveSkender: Nevermind.  You'd already caught on to these "no quotes" scenarios well.  Any number will do.

DaveSkender (approved):
All good.  Thanks for contributing! ❤️

PR #1099, opened 2023-10-12
fix: doc site button spacing

DaveSkender:
- [x] troubleshoot, fix, and test

- [ ] fix title line wrapping too


DaveSkender:
this line wrapping was from a browser Zoom, doesn't happen at normal mobile screen width

PR #1136, opened 2023-12-31
chore: Add semantic PR Linter

DaveSkender:
Adding enforcement of semantic [conventional commit standards](https://www.conventionalcommits.org/) for PR titles.

### to do

- [x] update settings, see [action docs](https://github.com/marketplace/actions/semantic-pull-request)
    - [x] enforce first letter capitalized, or sentence case
- [ ] ~~update PR template with other standards, like BREAKING CHANGE~~
- [ ] ~~consider also [Semantic Release convention](https://github.com/semantic-release/semantic-release) tooling~~
- [ ] ~~copy to other repos~~

> **Note**: to test in this initial PR, temporarily change `on: pull_request_target` to `on: pull_request`

### initial configuration

```
feat(ui): Add `Button` component
^    ^    ^
|    |    |__ Subject
|    |_______ Scope
|____________ Type
```
In our implementation:

- `types` and `scopes` requirements were not altered, using default settings and values (scope is not required, not do we recommend using it).
- `subject` was adjusted to required that it starts with a Capitalized letter, for example `feat: Poodle doodle` would pass, but `feat: poodle doodle` would fail.  This implicitly also means `subject` cannot start with symbols or numbers.


DaveSkender:
Why is it not triggered?

DaveSkender:
Okay, it works, just mot on this initial PR when using `on: pull_request_target`.

PR #1092, opened 2023-10-12
feat: Full chain streaming prototypes

DaveSkender:
### Description

As the next step in our [v3 preview release plan](https://github.com/DaveSkender/Stock.Indicators/discussions/1018), this implements:

- [x] refactor, simplify quote provider (observable only)
- stabilize full prototypes of the core _streaming_ scenarios
    - [x] #1198 
    - [x] quote/chain in » chain out - EMA,SMA
    - [x] #1214
    - [x] quote/chain in » unchainable result out - ALLIGATOR
    - [x] #1206 
- [x] #1230 for alligator, Renko (general, like EMA, SMA updates)
- [x] perf tuning when using `double.IsNaN()` over `null` evaluation (many places)
- [x] retry chainable type or interface over concrete typesto avoid hub type parameter, by de-coupling cache and observable types
   - [x] Remove `Reusable` type
   - [x] Remove `TQuote`, use `IQuote`
- [x] add back reference `Date` alias that writes to `long` timestamp, but `Date` marked as obsolete  but replace by complaint name
    - has constructors to take `DateTime` format
    - has _permanent_ name-compliant calculated property for `DateTime`
- [x] ~find and fix other, hidden fixed-period initialization indexes~ ([returned to backlog](https://github.com/users/DaveSkender/projects/1/views/1?pane=issue&itemId=72476662))
- [x] resolve commented issues (see comments section below)

### Problems to resolve

- [x] get full chain observation to work in isolated EMA/SMA base cases, proof of concept for #1096
- refactor to obviate dual cache, and other EMA/SMA performance tuning
    - [x] refactor performance tests to group `Static` or `Stream`, remove redundant series tests from stream grouping
    - [x] abstract ~`QuoteObserver`, `ChainObserver`, and~ generic `Cache` methods with generic `<T>` result storage
- [x] #1094 - add missing use cases for dispositions Acts, rebuild / cache busting scenarios
   - [x] or, at a minimum, address incomplete `OnNext()` scenarios and possible allow virtual overriding
- [x] allow non-chainable `IResult` to be an observable » all result forms should be observable, even if they're not chainable (for comprehensive end-use streaming).
- [x] consider: forcing concrete types for observables, cached quotes; I don't think `record struct` is going to work well with larger types, and has too many stack copies (stack overflow risks).  Allowing `TQuote` is only problematic on observing and caching; maybe just do a quick transform OnArrival(); 
   - » KEPT generic `TQuote`, found most issues were with index lookups; fixed in:
   - #1208 
[...]

PR #1150, opened 2024-01-26
feat: ToStringOut() for series

DaveSkender:
### Description

This is a quirky helper utility to convert `ISeries` (quotes and results lists) to a fixed width, CSV, or JSON string.  Mostly, I'm thinking this is helpful anyone debugging or wanting to output something to Console for quick visual inspections or general development sharing.  Default is `OutType.FixedWidth`, since I mostly had Console output and [.NET Fiddle](https://dotnetfiddle.net/xXJLuf) use in mind.

```console
   i  Timestamp      Open     High      Low    Close     Volume
---------------------------------------------------------------
   1  2017-01-03   212.71  213.350  211.520  212.570   96708880
   2  2017-01-04   212.64  214.220  212.638  213.648   83348752
   3  2017-01-05   213.14  214.060  213.020  213.685   82961968
   4  2017-01-06   213.41  215.170  213.414  214.318   75744152
   5  2017-01-09   213.87  214.530  213.866  214.193   49684316
   6  2017-01-10   214.03  214.890  213.520  214.083   67500792
```

### To do

- [x] depends on merge of #1092, since it restructures `IResult` to be parameterless
- [ ] question: isn't there a built-in set of converters that might work better here?  seems like we're re-engineering something that probably already exists in C# features.
- [ ] question: any way to use inner override of result class `ToString()`?  Probably not.  Maybe we just use `.ToStringList()` or `.ToStringOut()`
- [ ] fix: order date as first property, it's inherited, so it is always at the end right now
- [ ] fix: use of unused arguments for certain `OutType` values
- [ ] fix: formatting issues for date.  For `OutType.FixedWidth`, it would ideally be the smallest format possible, keeping in mind that bars could be high frequency date/time values or wider.
- [ ] consider overload `.ToString(int limitQty)` or `.ToString(int startIndex, int endIndex)`
- [ ] add separate backlog items if some classes might have their own (property, type, value, description (from XML docs)), ref: 
    ```console
    Property   Type          Value  Description
    -------------------------------------------------------------------
    Timestamp  DateTime 2024-02-02  The close date/time for this period
	Size       Decimal        3.00  High - Low price
    Body       Decimal        1.00  | Open - Close | (absolute value)
    UpperWick  Decimal        1.00  High - max(Open, Close)
    ...
    ```
- [ ] fix: overall performance, needs major refactoring
- [ ] add real unit tests


DaveSkender:
Note: overriding `ToString()` doesn't work on enumerable, but does work for base types.

Issue #1160, opened 2024-02-02
GetEma return NULL


MSLight:
This is my custom indicator class.

I applied the GetHeikinAshi function to my quotes, successfully obtaining Heikin-Ashi values. Subsequently, I aim to calculate the Exponential Moving Average (EMA) twice based on the close values derived from each Heikin-Ashi result. To achieve this, I utilized the GetEma function with a specific length parameter.

**However, I encountered an issue where the list of results contains NULL values for all instances of EMA, which is stored in the preEmaClose variable.**

If I directly calculate EMA on my quotes (without using GetHeikinAshi) it return result without problem.

How Can I solve the problem?

```
public static class CustomIndicator
    {
        public static List<ShaResult> GetSmoothedHa<TQuote>(
            this IEnumerable<TQuote> quotes,
            int preSmoothLength,
            int postSmoothLength)
            where TQuote : IQuote
        {
            var listQuotes = quotes.ToSortedCollection();
            List<ShaResult> results = new();

            var ha = listQuotes.GetHeikinAshi();

            var preEmaClose = haQuotes.GetEma(preSmoothLength);
            var postEmaClose = preEmaClose.GetEma(postSmoothLength);
            
            .
            .
            .

            return results;
        }
    }
```

DaveSkender:
@MSLight have you overcome your problems here?  If so, please close this issue or let us know how we can help.

PR #1161, opened 2024-02-02
fix: Heikin-Ashi docs

DaveSkender:
During evaluation of #1160 I found an error in our Heikin-Ashi documentation.  The first set of values is no longer `null` as there is a reasonable starting result based on the initial candle value.

To do:

- [x] remove reference that indicates the first period would be `null`
- [x] refactor code to only evaluate first period once.  This should provide a slightly better performance.


DaveSkender:
This micro-optimization for initialization improved the performance test by `0.001us` :neckbeard:  No difference, really.  Still, seems like a better practices so each iterative loop isn't re-evaluating condition.

PR #1200, opened 2024-06-01
docs: Add new "arrows" favicon

DaveSkender:
Still to do:

~- [ ] test/tweak formatting in dev build, then copy to to deploy also~
~- [ ] confirm `-1` syntax, we want to go back to prior tag, not prior commit in history~
~- [ ] see if _chore_ and _docs_ commits can be excluded~
~- [ ] post to build summary also, but without URL~
- [x] remove PR coverage comments but keep in build summary, just annoying (same in Py proj)
- [x] update favicons


.github/workflows/deploy-package.yml
```diff
@@ -55,12 +55,17 @@ jobs:
...
+      - name: Generate Release Notes
+        id: generate_release_notes
+        run: |
+          echo "## Release Notes" > release_notes.md
+          git log -1 --pretty=format:"%s" >> release_notes.md
+          echo >> release_notes.md
+          echo "[more on GitHub >>](https://github.com/DaveSkender/Stock.Indicators/releases/tag/${{ steps.compose.outputs.version }})" >> release_notes.md
+          echo "::set-output name=release_notes::$(cat release_notes.md)"
+
+      - name: Pack NuGet package
```
DaveSkender: remove, redundant task from copy/paste

DaveSkender:
this if far too fragile and unpredictable; and uneditable after published; reverting (bad idea)

PR #1235, opened 2024-07-31
refactor: Performance tuning (simplify caching analysis)

DaveSkender:
Undoing some over-engineered aspects of how we evaluate the arrival of observer items to determine caching action.  We were attempting to do a more graceful surgery to insert late arrivals; however, this ends up adding significant complexity and performance challenges.  This is an attempt to simplify to a more basic "add" or "rebuild from late arrival point" approach.

- [x] Remove excessive cache modification options
- [x] fix: cache overflow handler
- [x] Separate and simplify `IObservable` and `IObserver` interfaces
   - [x] refactor: class hierarchy
   - [x] refactor: move methods uniquely to `IObservable`, `IObserver`, `ICache` with less in integration `IHub` interface
   - [ ] refactor: use `in`/`out` interface type constraints to have better co/contravariance
- [ ] ~explore `ToIndicator()` as static method~ » can't do, relies on instance members
- [ ] ~explore putting static utilities directly in interface~ » no good opportunities


DaveSkender:
> [!IMPORTANT]
> Using vectorization works for time-series but has way too much instantiation overhead for hub (and likely incremental) approaches.  For hub, it went from 40us to 900us, but for time-series, it reduces the typical test case by 30-40% (from 14us to 10us) but is expected to have significantly more improvement with larger datasets and larger lookback windows.

DaveSkender:
Not much difference when comparing raw for loop versus vector maths on `double[]` equivalent prices.  Just for fun check.

| Method     | Mean     | Error      | StdDev    | Median   | Ratio | RatioSD |
|----------- |---------:|-----------:|----------:|---------:|------:|--------:|
| SmaArrLoop | 2.240 us |  0.3089 us | 0.0169 us | 2.231 us |  1.00 |    0.00 |
| SmaArrVect | 2.419 us | 15.9510 us | 0.8743 us | 1.921 us |  1.08 |    0.40 |

DaveSkender:
- [ ] Instead of ending transmission and cancellation subscriptions on fault, simply disable notification when faulted. This will allow resuming without re-subscription.

Issue #1091, opened 2023-10-12
add isolated increment methods


DaveSkender:

DaveSkender:
Since static increments are part of implementation streaming hubs, I’m repurposing this item to reflect full implementation of the _incremental_ approach mentioned in the last comment.  Here’s the original (obsolete scope) for reference:

> Implement incrementing scenarios to calculate the next indicator bar:
> 
> ### Internals
>
> - [ ] base increment[^1]
> - [ ] observer `OnNext()` handler
>
> ### Public APIs[^2]
> - [ ] base increment (system agnostic I/O), all needed inputs, appropriate `TResult` out
> - [ ] external cached (like observer handler, but using external data storage interface), consider [cache-aside](https://learn.microsoft.com/en-us/azure/architecture/patterns/cache-aside) design pattern.

[^1]: uses cached data instead of inputs

[^2]: repoints internal cache, then uses internal base increment?  this wouldn't perform well if provided data is re-written to memory, but am not sure how to handle CLR compliance otherwise with [by reference](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/method-parameters#ref-parameter-modifier).

codebeaulieu:
This is awesome, had this ideas in mind a while back and never got around to implementing it. Great work Dave.

Issue #1062, opened 2023-05-20
Incrementally update


ingted:
It seems like the indicator values are calculate for the whole input.

Is there any possible to add incrementally update version?

Here are my workaround for the Adx indicator:

Adx.Api.cs
```
public static partial class Indicator
{

    public static Tuple<List<AdxResult>, AdxHelper> GetAdx<TQuote>(
        this IEnumerable<TQuote> quotes,
        AdxHelper ah,
        List<AdxResult> results,
        int start,
        int lookbackPeriods = 14
        )
        where TQuote : IQuote
    {
        return quotes
            .ToQuoteD()
            .CalcAdx(lookbackPeriods, ah, results, start);
    }
}
```

AdxSeries.cs
```
namespace Skender.Stock.Indicators;


public class AdxHelper
{
    public double prevHigh = 0;
    public double prevLow = 0;
    public double prevClose = 0;
    public double prevTrs = 0; // smoothed
    public double prevPdm = 0;
    public double prevMdm = 0;
    public double prevAdx = 0;

    public double sumTr = 0;
    public double sumPdm = 0;
    public double sumMdm = 0;
    public double sumDx = 0;
}
// AVERAGE DIRECTIONAL INDEX (SERIES)
public static partial class Indicator
{
    internal static Tuple<List<AdxResult>, AdxHelper> CalcAdx(
        this List<QuoteD> qdList,
        int lookbackPeriods,
        AdxHelper ah,
        List<AdxResult> results,
        int start = 0)
    {
        // check parameter arguments
        ValidateAdx(lookbackPeriods);

        // initialize
        int length = qdList.Count;
        if (results == null)
        {
            results = new(length);
        }

        //double prevHigh = 0;
        //double prevLow = 0;
        //double prevClose = 0;
        //double prevTrs = 0; // smoothed
        //double prevPdm = 0;
        //double prevMdm = 0;
        //double prevAdx = 0;

        //double sumTr = 0;
        //double sumPdm = 0;
        //double sumMdm = 0;
        //double sumDx = 0;
        if (ah == null)
        {
            ah = new AdxHelper();
        }
        // roll through quotes
        for (int i = start; i < length; i++)
        {
            QuoteD q = qdList[i];

            AdxResult r = new(q.Date);
            results.Add(r);

            // skip first period
            if (i == 0)
            {
                ah.prevHigh = q.High;
                ah.prevLow = q.Low;
                ah.prevClose = q.Close;
                continue;
            }

[...]

PR #1271, opened 2024-11-11
feat: Buffer-style ADX incremental

DaveSkender:
Implements #1062.  This behaves just like a normal `List()` you add quotes incrementally.  Under the hood, it'll convert that to an incremental `AdxResult` and add it to the end. `AdxList` is equivalent to `List<AdxResult>`.

```csharp
AdxList adxList = new(lookbackPeriods: 14);

foreach (IQuote quote in quotes)
{
    adxList.Add(quote);
}
```

This buffer-style list is a lightweight, simple way to implement incremental indicators.  It's a middle alternative between the performance of time-series batches (existing v2 library) and the more advanced _streaming hub_ ecosystem.

Done when

- [ ] ~refactor to `abstract` `BufferList():List`~
- [x] shorten path from `IEnumerable` to date/value `Add()`
- [x] add ADX buffer-style list

Related to:
- #1091


DaveSkender:
There's some refactoring still do do here, which I plan to do when implementing the hub.

Here's some relative performance between this buffer list and the straight time series.  For comparison, I've included the notably simpler calculating EMA for comparison to the _streaming_ approach.  Generally, the incrementing buffer approach has more overhead and is 2-3× less performant than the full _original_ static time series.  The streaming hub is slower than both, but is more robust with larger datasets.

| Method    | Mean      | Error      | StdDev    |
|---------- |----------:|-----------:|----------:|
| AdxBuffer | 51.621 us | 11.4208 us | 0.6260 us |
| AdxSeries | 14.169 us |  2.7392 us | 0.1501 us |
| EmaBuffer |  4.019 us |  0.7001 us | 0.0384 us |
| EmaSeries |  2.621 us |  0.6898 us | 0.0378 us |
| EmaStream | 27.899 us | 16.6569 us | 0.9130 us |

PR #1327, opened 2025-04-15
chore: Cleanup dev container

DaveSkender:
#### done when

- [x] Remove unwanted extensions and settings


DaveSkender:
Codacy is offline

PR #1595, opened 2025-10-23
style: Roslynator "hidden" diag-level cleanup

DaveSkender:
What happens when you run Roslynator at the "hidden" sub-info level.
DO NOT MERGE.  Experiment only, based on:

- #1594


DaveSkender:
@coderabbitai full review but only spot check core files, do not abort due to file count, pick the core items only.

PR #1740, opened 2025-11-29
feat: Add Volume Profile indicator (update)

JGronholz:
Includes fix to cumulative dictionary (per-result copy).


DaveSkender:
❤️ thanks Jon — will take a look when I have some time to run through some manual calculations to wrap my brain around this.

JGronholz:
This attributes all the volume in a candle in the direction of Open->Close, which accumulates error.
To minimize the magnitude of the error, use the largest quantity of lowest-timeframe candles you can afford to.

This chops each candle's price action into equally-sized buckets.
If Close > Open, Volume is added evenly to the buckets.
If Open > Close, Volume is subtracted evenly from the buckets.
This is the underlying simplicity of it.

The mechanism to propagate the buckets forward without:
1) accumulating rounding errors
2) excessive copying

was tricky.
This is the complexity wrapped around the simplicity.

To profile a specific timespan:
Candles.Where(startDate < timestamp < endDate).VolumeProfile

Visible range (VRVP):
Candles.Where(viewStart < timestamp < viewEnd).VolumeProfile

PR #1819, opened 2025-12-31
test: Add missing regression test stubs

DaveSkender:


DaveSkender:
@copilot Renko ATR regression tests (`tests/indicators/m-r/RenkoAtr/RenkoAtr.Regression.Tests.cs`) were missing `[TestMethod]` attributes and are not implemented.  I've added the attributes for this one test class and inconclusive asserts as placeholders.  Check for and update all such corrections for similar regression tests so I can easily identify them and add tests later.  Do not implement any tests, we're just identifying them consistently in this manner for visibility.

PR #1822, opened 2025-12-31
docs: XML comments cleanup

DaveSkender:


DaveSkender:
@copilot Update all `StreamHub` implementations' XML docs using `EmaHub` (`src/e-k/Ema/Ema.StreamHub.cs`) as a reference copy.

Be meticulously diligent about implementing with the same phrasing and intent.  For example:

- use careful consideration for how they're written and sequence of phrasing in descriptors
- use `<inheritdoc />` where possible if a base has already written the XML descriptors, avoid duplication
- avoid adding verbose XML docs on non-public interfaces, or exclude entirely if repetitive or self-evident (self documenting names)
- the extension methods like `ToEmaHub` are where we have the more thoughtful public interfaces, as such, the `internal` constructors do not need any XML docs at all.
- top level class descriptors should be succinct and in the same phraseology as shown

PR #1837, opened 2026-01-03
feat: Add StreamHubState<TIn, TState, TOut> for stateful indicators with cached rollback

JGronholz:
The only changes to StreamHub are the "virtual" modifiers.
I added the stateful variant (StreamHubState) to avoid breaking changes.
For example, I also implemented the stateful rsi hub: RsiHubState.

Although this doesn't make RollbackState an O(1), it restores the state, rather than reconstructing it, which normalizes the compute cost across all indicators.


src/m-r/Rsi/Rsi.StreamHubState.cs
```diff
@@ -0,0 +1,207 @@
+namespace Skender.Stock.Indicators;
+
+/// <summary>
+/// State object for RSI streaming hub.
+/// </summary>
+/// <param name="AvgGain">Average gain over lookback period.</param>
+/// <param name="AvgLoss">Average loss over lookback period.</param>
+public record RsiState(double AvgGain, double AvgLoss) : IHubState;
```
DaveSkender, line 8: This is similar to what I'm doing in the `BufferList` style, except I'm using it there for a rolling cache.  This is a solid idea, especially for RSI where there's potentially redundant recalculations on rollback.

https://github.com/DaveSkender/Stock.Indicators/blob/657f3f597b27df4742b842f1792bea1ab0856dec/src/m-r/Rsi/Rsi.BufferList.cs#L8

DaveSkender (commented):
This is a solid idea and you're 100% right in spotting an area where some performance tuning is certainly needed.

The nuance with _performance_ though is that it's not always intuitive which tradeoffs are actually better.  I think with this example, you're trading recalculation for I/O storage operations, which end up being slower.  With that said, I think there might be a way to make this work by avoiding some of the copying that happens?

```bash
dotnet run --project tools/performance -c Release --filter *StreamIndicators.RsiHub*
```

| Method      | Mean     | Error    | StdDev   |
|------------ |---------:|---------:|---------:|
| RsiHub      | 31.22 us | 31.62 us | 1.733 us |
| RsiHubState | 49.80 us | 22.73 us | 1.246 us |

```bash
dotnet run --project tools/performance -c Release --filter *StreamIndicators.SmaHub*
```

| Method      | Mean     | Error    | StdDev   |
|------------ |---------:|---------:|---------:|
| SmaHub      | 35.39 us | 14.95 us | 0.819 us |
| SmaHubState | 63.19 us | 35.18 us | 1.928 us |

```bash
dotnet run --project tools/performance -c Release --filter *StreamIndicators.StdDevHub*
```

| Method         | Mean     | Error    | StdDev   |
|--------------- |---------:|---------:|---------:|
| StdDevHub      | 63.68 us | 12.90 us | 0.707 us |
| StdDevHubState | 99.45 us | 27.60 us | 1.513 us |

JGronholz:
@DaveSkender, there was no benchmark for rapid updates to the latest candle, so no baseline existed for the specific issue this branch solves. Copilot is collecting those baselines now and will update docs\RAPID-UPDATE-BENCHMARKS.md. Then, I'll finish collecting the hubstate benchmarks, and we'll see where we're at. It's looking like this will make 1/3 much faster, 1/3 approx. the same +/-, and 1/3 worse, so we'll need to implement this strategically.

Adding the "virtuals" to streamhub made this experiment possible. If you committed _just that_, it would allow me to relocate my derived hubstates, and return my v3 to canonical. I'm using the prerelease directly in my app, and don't mind contaminating my app instead of the library, to test. ❤️

DaveSkender:
> there was no benchmark for rapid updates to the latest candle, so no baseline existed for the specific issue this branch solves. Copilot is collecting those baselines now and will update docs\RAPID-UPDATE-BENCHMARKS.md.

Sounds good, it'll be interesting to see what you come up with.  I've also been working on updating the performance test suite and will take some of your ideas and try to implement them.

- #1848 

> Then, I'll finish collecting the hubstate benchmarks, and we'll see where we're at. It's looking like this will make 1/3 much faster, 1/3 approx. the same +/-, and 1/3 worse, so we'll need to implement this strategically.

Yah, it's never a one-solution thing with some of these since they tend to have different complexities.  In that PR (#1848) branch, I've updated the baselines and added tasks for some of the [performance hot spots](https://github.com/DaveSkender/Stock.Indicators/blob/update-perf-test-suite/docs/plans/streaming-indicators.plan.md#critical-streamhub-performance-issues) and will take a look at those soon alongside what you've come up with here.

> Adding the "virtuals" to streamhub made this experiment possible. If you committed _just that_, it would allow me to relocate my derived hubstates, and return my v3 to canonical. I'm using the prerelease directly in my app, and don't mind contaminating my app instead of the library, to test. ❤️

I'm not opposed to making things `virtual` in certain cases, I've done that in many places.  Something to keep in mind though is that you can internalize some of these caches you're implementing in the `StreamHub` base class too, rather than externalizing them (may not need `virtual` in some cases?).

JGronholz:
Let's put this on hold, for now, pending #1878

PR #1864, opened 2026-01-05
feat: Extend `Use()` adapter

DaveSkender:
Expenses the `Use` extension utility to handle other ISeries types of indicator results that don’t have a natural.  Currently `quotes.Use(…)` only supports converting `IQuote` list types to `QuotePart` types.  This implements `series.Use(“my-property-name”)` for wider use.  We may need to rename `QuotePart` to `Reuseable`.

Related to:
- #1865 
- #1866 
- #1878

### DIFF
--- a/src/_common/Candles/Candles.Utilities.cs
+++ b/src/_common/Candles/Candles.Utilities.cs
@@ -6,6 +6,28 @@ namespace Skender.Stock.Indicators;
 public static class Candlesticks
 {
     /// <summary>
+    /// Converts Candle results to a chainable list using the Price field.
+    /// </summary>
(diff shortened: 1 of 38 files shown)


JGronholz:
`series.Use(“my-property-name”)`

Are you already working on the UseHub? 😁👍

DaveSkender:
> `series.Use(“my-property-name”)`
> 
> Are you already working on the UseHub? 😁👍

Mostly just seeing what this might look like as an alternative to trying to force IReusable with NaN on non-chainable types.  I don’t love this approach because it’s very flakey with a lot of overhead and am still not entirely sure this is a problem worth solving yet.

PR #1866, opened 2026-01-05
docs: Result classes implementing ISeries

JGronholz:
I'm loading indicators dynamically, and must operate via interface alone.
IStreamHub isn't out variant, so I cannot make strongly-typed piles of hubs.
I'd like to make IStreamHub implement a base interface with out variance.
That would allow me to make piles of `IStreamHubBase<ISeries>`, which is basically useless... since I'd only have access to Timestamp without casting.
However, if all results implemented IReusable, I could make piles of `IStreamHubBase<IReusable>`, and quickly iterate (dictionary key, LastItem/Results (Value!!!)) tuples, without knowing anything about the hubs.
If Value isn't what I'm after (or it's null), I can reflect/cast/is/switch the result.

Some of the ISeries results have a single, obvious, value, and some don't.
What about:
1) Make them all IReusable
2) Override the ones that don't currently have Value to return null
Should be non-breaking

### DIFF
--- a/ISeries-vs-IReusable-Analysis.md
+++ b/ISeries-vs-IReusable-Analysis.md
@@ -1,0 +1,520 @@
+# Result classes analysis: ISeries vs IReusable
+
+This document lists all result classes that implement `ISeries` instead of `IReusable`, along with recommended properties for the `Value` property if they were to implement `IReusable`.
+
+## Executive summary
+
+- **Total result classes implementing ISeries (not IReusable):** 16
+- **Total result classes implementing IReusable:** 63
+- **CandleResult:** Also implements ISeries (1 additional class)
+
+## Understanding the interfaces
+
+### ISeries interface
+
+```csharp
+public interface ISeries
+{
+    DateTime Timestamp { get; }
+}
+```
+
+The `ISeries` interface only requires a `Timestamp` property and is the base interface for all time-series data.
+
+### IReusable interface
+
+```csharp
+public interface IReusable : ISeries
+{
+    [JsonIgnore]
+    double Value { get; }
+}
+```
+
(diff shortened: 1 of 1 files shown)


DaveSkender (commented):
> Some of the ISeries results have a single, obvious, value, and some don't.
> What about:
> 
> Make them all IReusable
> Override the ones that don't currently have Value to return null Should be non-breaking

Returning `null` would be a large performance reduction throughout as it forces null evaluation into all maths.

Forcing `IReusable` on to all and just returning virtual double.NaN in results class is feasible, but will lead to an unclear interface.  I’ll have to think about this more.

I see 2-3 of these that make sense to make IReusable, like AtrStop.

Try
1. Use piles of StreamHub<T> where T: IReusable May workaround here, guessing
2. Check out new Catalog features in lib — made for this kind of automation of inventory.

JGronholz:
I appreciate your consistency! The fail-safe approach says if it _could_ happen, it will...

Realistically, though, if Ichimoku did have a Value, what would you chain it to?
Likewise, what indicator would Supertrend chain to?
The real issue is that some of these indicators have no business being chained.
Not providing a Value for those indicators does makes logical sense, but creates this inheritance/variance dilemma (for me lol).
I didn't pay attention and thought Value was already nullable; NaN works for me!
Since the Value property doesn't exist for these indicators, this is non-breaking.
I'm not even sure a disclaimer is warranted.
If someone wants to take the RSI of the EMA of the Ichimoku's Senku... let them???
I assume Slope is used downstream of other indicators more than any other, but maybe I'm wrong.

```
public interface IStreamHubBase<out TOut>
    where TOut : ISeries
{
   IReadOnlyList<TOut> Results { get; }
}

public interface IStreamHub<in TIn, TOut> : IStreamHubBase<TOut>
    where TIn : ISeries
    where TOut : ISeries

```
This adds the out variance, allowing piles of `IStreamHubBase<ISeries>`, without changing anything else.
Now, I just need to swap ISeries for IReusable and I'm set...
Except the ISeries results don't fit the constraint.
This is what creates my need to homogenize the results interface.

PR #1878, opened 2026-01-05
feat: Add out variance to IStreamHub

JGronholz:
This PR:
1) Adds virtuals to the concrete type (allows state-caching experiments)
2) Does NOT include the StreamHubState experiment ^
3) Adds out variance to IStreamHub via the IStreamHubBase interface
4) Changes the constraints from ISeries to IReusable
5) All results now implement IReusable. Modified result classes return double.NaN for Value. (including CandleResult)

No build errors.
All hubs can now be cast to IStreamHubBase<IReusable>.
Value can be retrieved without knowing anything about the hub.

This makes retrieval for heuristics O(minimal).
```
Dictionary<(symbol, interval, name, paramsHash) hubkey, IStreamHubBase<IReusable> hub> _hubs;
_hubs[hubkey].Results.LastOrDefault().Value;
```

This works!!!

### DIFF
--- a/src/_common/Candles/CandleResult.cs
+++ b/src/_common/Candles/CandleResult.cs
@@ -4,7 +4,7 @@ namespace Skender.Stock.Indicators;
 /// Represents the result of a candlestick analysis.
 /// </summary>
 [Serializable]
-public record CandleResult : ISeries
+public record CandleResult : IReusable
 {
     /// <summary>
     /// Initializes a new instance of the <see cref="CandleResult"/> record.
@@ -64,4 +64,6 @@ public record CandleResult : ISeries
     /// Gets the candlestick properties.
     /// </summary>
     public CandleProperties Candle { get; init; }
+
+    public double Value => double.NaN;
 }
--- a/src/_common/StreamHub/IStreamHub.cs
+++ b/src/_common/StreamHub/IStreamHub.cs
@@ -1,14 +1,5 @@
 namespace Skender.Stock.Indicators;
 
-public interface IStreamHubBase<out TOut>
-    where TOut : IReusable
-{
-    /// <summary>
-    /// Read-only list of the stored cache values.
-    /// </summary>
-    IReadOnlyList<TOut> Results { get; }
-}
-
 /// <summary>
 /// Streaming hub: management of observer
 /// and observable indicator data
--- a/src/_common/StreamHub/IStreamHub.cs
+++ b/src/_common/StreamHub/IStreamHub.cs
@@ -20,7 +20,7 @@ public interface IStreamHubBase<out TOut>
 /// Type of outbound indicator data.
 /// </typeparam>
 public interface IStreamHub<in TIn, TOut> : IStreamHubBase<TOut>
-    where TIn : ISeries
+    where TIn : IReusable
     where TOut : IReusable
 {
     /// <summary>
--- a/src/_common/StreamHub/IStreamHub.cs
+++ b/src/_common/StreamHub/IStreamHub.cs
@@ -1,7 +1,7 @@
 namespace Skender.Stock.Indicators;
 
 public interface IStreamHubBase<out TOut>
-    where TOut : ISeries
+    where TOut : IReusable
 {
     /// <summary>
     /// Read-only list of the stored cache values.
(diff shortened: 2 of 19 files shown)


JGronholz:
1) Little bit of XML comments to update.
2) Should Value have [jsonignore]?

DaveSkender:
I can see what you’re trying to do by having a common IStreamHubBase usable for whatever generic `Value` evaluations you had in mind.  I like giving ability to have `IStreamHub` uses like this, but am still trying to figure how confusing this’ll be for indicators with unusable values.

Aside from the basic ability to do C# type stacking, what’s the functional benefit?  I mean, all `Value` outputs aren’t fungible in most cases, right?

JGronholz:
The functional benefit is to downstream heuristics:
```
If (hubs["RSI_4h"].LastValue > 70) return Overbought(4h);

If (hubs["EMA_Fast_4h"].LastValue > hubs["EMA_Slow_4h"].LastValue) return MomentumUp(4h);

If ("EMA_Fast_TF" > "EMA_Mid_TF" > "EMA_Slow_TF") return EmaStackAlignment(TF, Up);
```
for example...

All my hubs are keyed by (symbol, interval, etc.).
If the downstream consumer didn't already know what Value will mean, and what to do with it, it wouldn't be able to generate the lookup key in the first place.
This allows dynamically-loaded virtual heuristic descriptors (json-constructed) to use dynamically-loaded virtual indicator descriptors (json-constructed). (virtual=not hard-wired, factories returning object?, etc.)
Indexing was half of the solution; IReusable.Value was the other.
The dictionary key is the "context" for each hub. If you know one, you know the other.

JGronholz:
Representative side-effect of this change:

"So the Value property was added in commit 4c69db6 "Swap ISeries for IReusable". The IReusable interface requires a Value property, so when AlligatorResult was changed from ISeries to IReusable, it needed to add the Value property. Since Alligator has three separate lines (Jaw, Teeth, Lips) and no single "value", it returns NaN.

The Condense test is now failing because it removes all items with NaN values. This test needs to be updated to reflect the new reality that AlligatorResult doesn't have a meaningful single Value property.

Let me fix the test to check for results where all three properties (Jaw, Teeth, Lips) are null instead of checking the Value property:"

This needs a comprehensive sweep for such side-effects (errors probably localized to tests).

DaveSkender:
Have you tried something like:


```csharp
QuoteHub quoteHub = new();

List<IStreamHub<IQuote,IReusable>> hubs = [
  ..quoteHub.ToEmaHub(14),
  ..quoteHub.ToEmaHub(20),
  ..
];
``` 

It might be easier to make some kind of `HubCollection` class?

---

With that said, for your use case, here’s how I expect to do your EMA crossover:

```csharp
QuoteHub quoteHub = new();

EmaHub emaFast = quoteHub.ToEmaHub(50);
EmaHub emaSlow = quoteHub.ToEmaHub(200);

// add quotes to quoteHub
// and the 2 EmaHub will be in sync

if(emaFast.Results[^2].Ema < emaSlow.Results[^2].Ema  // or .Value
&& emaFast.Results[^1].Ema > emaSlow.Results[^1].Ema)
{
    // cross over occurred
}
```

JGronholz:
This works flawlessly, hard-wired:
```
QuoteHub quoteHub = new();
EmaHub emaFast...
```

My QuoteHub factory method always returns a QuoteHub, so this was no problem.
My IndicatorHub factory method(s) return types derived from IStreamHub.

`IStreamHub<IQuote, IReusable>` was the key. Thank you!

I must have tried every combination... except this one. Or the constraint was changed in a recent commit? I'm not even sure how I became convinced I needed out variance, anymore, since IReusable was always ok with me. More confused than when I started, but the current v3 works, as-is, so all my experiments can be abandoned.

Back on track, the perf of RollBackState() is the only "bottleneck" for me (solid perf, but my longest hot-path).
I will use the hubs as synchronous IReusable extruders.
After each Add(quote)/close, persist LastItem from all quote and indicators hubs.
Leaves hub cache isolated (by duplicating it).

The dictionary keys give context to IReusable.Value:
`private readonly ConcurrentDictionary<(string Symbol, KlineInterval Interval, string IndicatorKey, string ParamsHash), IStreamHub<IQuote, IReusable>> _hubs = new();`
IndicatorKey = "Ema", ParamsHash = "5" //Ema_5
IndicatorKey = "Ema_5" //lookback baked into descriptor/config
IndicatorKey = "Ema_Short" //lookback baked into descriptor/config

PR #1894, opened 2026-01-10
feat: Add extensible ReusableObserver implementing IStreamObserver<IReusable>

JGronholz:
Everything I was trying to gain access to by modifying the base interfaces (observables) was already available at the other end (observer).
This class generically implements a specific run-time configuration of `IStreamObserver<IReusable>`.

TODO: It could probably be made generic to provide type-safety to hard-wired instances.

### DIFF
--- a/src/_common/StreamHub/ReusableObserver.cs
+++ b/src/_common/StreamHub/ReusableObserver.cs
@@ -6,6 +6,7 @@ namespace Skender.Stock.Indicators;
 /// </summary>
 public class ReusableObserver : IStreamObserver<IReusable>
 {
+    private readonly Func<bool> _isSubscribed;
     private readonly Action<Exception>? _onError;
     private readonly Action? _onCompleted;
     private readonly Action? _onUnsubscribe;
@@ -16,11 +17,11 @@ public class ReusableObserver : IStreamObserver<IReusable>
     private readonly Action? _rebuild;
     private readonly Action<DateTime>? _rebuildTimestamp;
     private readonly Action<int>? _rebuildIndex;
-    private bool _isSubscribed;
 
-    public bool IsSubscribed => _isSubscribed;
+    public bool IsSubscribed => _isSubscribed();
 
     public ReusableObserver(
+        Func<bool> isSubscribed,
         Action<Exception>? onError = null,
         Action? onCompleted = null,
         Action? onUnsubscribe = null,
@@ -32,6 +33,7 @@ public class ReusableObserver : IStreamObserver<IReusable>
         Action<DateTime>? rebuildTimestamp = null,
         Action<int>? rebuildIndex = null)
     {
+        _isSubscribed = isSubscribed;
         _onError = onError;
         _onCompleted = onCompleted;
         _onUnsubscribe = onUnsubscribe;
@@ -48,7 +50,7 @@ public class ReusableObserver : IStreamObserver<IReusable>
 
     public void OnCompleted() => _onCompleted?.Invoke();
 
-    public void Unsubscribe() { _isSubscribed = false; _onUnsubscribe?.Invoke(); }
+    public void Unsubscribe() { _onUnsubscribe?.Invoke(); }
 
     // OnAdd is called when a new item is added to the hub's cache
     public void OnAdd(IReusable item, bool notify, int? indexHint) => _onAdd?.Invoke(item, notify, indexHint);
@@ -60,7 +62,7 @@ public class ReusableObserver : IStreamObserver<IReusable>
     public void OnPrune(DateTime toTimestamp) => _onPrune?.Invoke(toTimestamp);
 
     // Reinitialize is called to reset the observer state
-    public void Reinitialize() { _isSubscribed = true; _onReinitialize?.Invoke(); }
+    public void Reinitialize() { _onReinitialize?.Invoke(); }
 
     // Rebuild methods trigger recalculation
     public void Rebuild() => _rebuild?.Invoke();
--- a/src/_common/StreamHub/ReusableObserver.cs
+++ b/src/_common/StreamHub/ReusableObserver.cs
@@ -6,7 +6,6 @@ namespace Skender.Stock.Indicators;
 /// </summary>
 public class ReusableObserver : IStreamObserver<IReusable>
 {
-    private readonly Action<IReusable>? _onNext;
     private readonly Action<Exception>? _onError;
     private readonly Action? _onCompleted;
     private readonly Action? _onUnsubscribe;
@@ -22,7 +21,6 @@ public class ReusableObserver : IStreamObserver<IReusable>
     public bool IsSubscribed => _isSubscribed;
 
     public ReusableObserver(
-        Action<IReusable>? onNext,
         Action<Exception>? onError = null,
         Action? onCompleted = null,
         Action? onUnsubscribe = null,
@@ -34,7 +32,6 @@ public class ReusableObserver : IStreamObserver<IReusable>
         Action<DateTime>? rebuildTimestamp = null,
         Action<int>? rebuildIndex = null)
     {
-        _onNext = onNext;
         _onError = onError;
         _onCompleted = onCompleted;
         _onUnsubscribe = onUnsubscribe;
@@ -46,8 +43,6 @@ public class ReusableObserver : IStreamObserver<IReusable>
         _rebuildTimestamp = rebuildTimestamp;
         _rebuildIndex = rebuildIndex;
     }
-
-    public void OnNext(IReusable value) { _onNext?.Invoke(value); }
 
     public void OnError(Exception exception) => _onError?.Invoke(exception);
 
--- a/src/_common/StreamHub/ReusableObserver.cs
+++ b/src/_common/StreamHub/ReusableObserver.cs
@@ -1,0 +1,76 @@
+namespace Skender.Stock.Indicators;
+
+/// <summary>
+/// Flexible IStreamObserver implementation with customizable callbacks for all lifecycle methods.
+/// Most callbacks are optional (nullable) except onNext which is required.
+/// </summary>
+public class ReusableObserver : IStreamObserver<IReusable>
+{
+    private readonly Action<IReusable> _onNext;
+    private readonly Action<Exception>? _onError;
+    private readonly Action? _onCompleted;
+    private readonly Action? _onUnsubscribe;
+    private readonly Action<IReusable, bool, int?>? _onAdd;
+    private readonly Action<DateTime>? _onRebuildFromTimestamp;
+    private readonly Action<DateTime>? _onPrune;
+    private readonly Action? _onReinitialize;
+    private readonly Action? _onRebuild;
+    private readonly Action<DateTime>? _onRebuildTimestamp;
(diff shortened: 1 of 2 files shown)


JGronholz:
1) Not sure where OnNext came from... OnAdd was originally routed to onNext... CopilotD10T error fixed.
2) All IStreamObserver members route directly to corresponding callbacks, one each, including IsSubscribed

JGronholz:
If you don't consider this useless, it still needs proper integration into the library (catalog?/etc).

Example usage:
```
var observer = new ReusableObserver(
    isSubscribed: () => true,
    onError: error =>
    {
        _logger.LogError(error, "Error in hub subscription for {Key}", key);
    },
    onCompleted: () =>
    {
        _logger.LogInformation("Hub subscription completed for {Key}", key);
    },
    onAdd: (item, notify, indexHint) =>
    {
        if (notify)
        {
            //fire-and-forget; don't await in hot-path
            //_ = _persistence.Upsert(item);

            // Store the latest result
            _lastResults[key] = item;

            // Update UI on dispatcher thread
            //async invoke; don't await in hot-path
            //TODO: throttle UI updates
            Application.Current?.Dispatcher.BeginInvoke(() =>
            {
                UpdateIndicatorValue(symbol, interval, indicatorName, paramsHash, item, key);
            }, System.Windows.Threading.DispatcherPriority.Background);
        }
    });
```

JGronholz:
This is basically the observer third of StreamHub:
```
// STREAM HUB (OBSERVER)
public abstract partial class StreamHub<TIn, TOut> : IStreamObserver<TIn>
```
except not hard-coded to do anything specific.

We can't produce observable TOut's without first observing TIn's.
If `ReusableObserver` implemented `IStreamObserver<ISeries>` instead of `<IReusable>`, `StreamHub<TIn, TOut>` could probably derive from it.
That would allows buckets of `ReusableObserver`'s (same goal as `ChainHub<IQuote, IReusable>`, except wouldn't have access to IReusable.Value).

I like the idea of making `ReusableObserver` generic, but don't want to introduce variance issues.
`DonchianObserver : ReusableObserver<DonchianResult>` would have full access to the DonchianResult.
This would allow each ISeries (and IReusable) use-case to be individually handled by the end-user:
```
DonchianCenterlineObserver : ReusableObserver<DonchianResult> { ... }
DonchianWidthObserver : ReusableObserver<DonchianResult> { ... }
```

JGronholz:
Latest commit added the generic!
ReusableObserver works for all existing result types (ISeries constraint).
Rather than adding Value=NaN to results that can't be chained in a straightforward way, this compels the end-user to explicitly implement onAdd<TResult>.
```
          DonchianObserver : ReusableObserver<DonchianResult> { ... }
DonchianCenterlineObserver : ReusableObserver<DonchianResult> { ... }
     DonchianWidthObserver : ReusableObserver<DonchianResult> { ... }
```

DaveSkender:
If I’m interpreting your intent correctly, you want:

1. to use lambda functions instead of normal method implementation for `IObservable<T>` classes outside of the library — this looks to just enable that aspect, but doesn’t really make anything easier for users (still have to write the method implements).
2. ultimately, to denormalize all indicator out values into `Value` types (for unclear reasons)

To be honest, I’m struggling to see the value in either of these (pun intended).  I’d probably need to see more of what this would improve outside the library to truly see it (e.g. before this vs. with this).

JGronholz:
I challenged copilot to put an RsiHub and MacdHub in the same List<> and IChainProvider<IReusable> was THE shared interface it found. Everything else doesn't work due to invariance in the interfaces.
```
var provider = new QuoteHub();
var hubs = new List<IChainProvider<IReusable>>();

var rsiHub = provider.ToRsiHub(14);
hubs.Add(rsiHub);

var macdHub = provider.ToMacdHub(12, 26, 9);
hubs.Add(macdHub);

var value = hubs[0].Value;
```
This does work, and gets me access to Value via `IChainProvider:IStreamObservable`, without knowing anything about the concrete types. Asdf() example uses a List<>, but reality is a `Dictionary<key, IStreamObservable<IReusable>>`.
`_dictionary["Rsi_14"].Value`

So, now I can repeatedly poll all my hubs and copy Value somewhere...
Except, yuck!

If only there was some kind of universal thing I could subscribe to the ends of all my hub chains to actually *DO* something with the calculated indicator value(s)... invoke UI callbacks... draw candles... persist... hello world... anything.

Quotes > Indicators > X

*I* am struggling to figure out what X is, for you.
For me, it's heuristics/etc...
For me, the hubs are worthless if they don't produce a Value.
Only 16 hubs/results aren't IReusable, so this isn't a deal-breaker causing me to entirely abandon v3.

I think I'm on the right track---

Current option:
-My persistence service implements `IStreamObserver<TResult>` and subscribes to every "leaf" (end-of-line) hub.

New option:
-I inject my persistence callback into a (reusable) ReusableObserver<TResult> instance and subscribe IT to every "leaf" hub. My persistence service creates one for its' purpose (persistence), and my viewmodels create them for their specific purposes (charting, datagrid, etc.).

Datagrid only cares about Results.Last().Value:
```
var observer = new ReusableObserver<IReusable>(
    isSubscribed: () => true,
    onError: error =>
    {
        _logger.LogError(error, "Error in hub subscription for {Key}", key);
    },
    onCompleted: () =>
    {
        _logger.LogInformation("Hub subscription completed for {Key}", key);
    },
    rebuild: () =>
    {
        var item = hub.GetCacheRef().LastOrDefault();
        if (item != null)
            updateIndicatorValue(item, symbol, interval, indicatorName, paramsHash, key);
    },
    onRebuild: (fromTimestamp) =>
    {
        var item = hub.GetCacheRef().LastOrDefault();
        if (item != null)
            updateIndicatorValue(item, symbol, interval, indicatorName, paramsHash, key);
    },
    rebuildIndex: (fromIndex) =>
    {
        var item = hub.GetCacheRef().LastOrDefault();
        if (item != null)
            updateIndicatorValue(item, symbol, interval, indicatorName, paramsHash, key);
    },
    rebuildTimestamp: (fromTimestamp) =>
    {
        var item = hub.GetCacheRef().LastOrDefault();
        if (item != null)
            updateIndicatorValue(item, symbol, interval, indicatorName, paramsHash, key);
    },
    onAdd: (item, notify, indexHint) =>
    {
        if (notify)
        {
            updateIndicatorValue(item, symbol, interval, indicatorName, paramsHash, key);
        }
    });

private void updateIndicatorValue(IReusable item, string symbol, KlineInterval interval,
        string indicatorName, string paramsHash, string key)
{
    _lastResults[key] = item;

    // Update UI on dispatcher thread
    Application.Current?.Dispatcher.BeginInvoke(() =>
    {
        UpdateIndicatorValue(symbol, interval, indicatorName, paramsHash, item, key);
    }, System.Windows.Threading.DispatcherPriority.Background);
}

private void UpdateIndicatorValue(string symbol, KlineInterval interval, 
        string indicatorName, string paramsHash, IReusable result, string key)
{
    try
    {
        // Find the item in the collection
        var item = IndicatorValues.FirstOrDefault(i =>
            i.Symbol == symbol && i.Interval == interval.ToString() &&
            i.IndicatorType == indicatorName && i.SeriesKey == paramsHash);

        if (item != null)
        {
            // Update existing item
            item.Value = result?.Value ?? 0.0;
            item.BarTimestamp = result?.Timestamp ?? DateTime.UtcNow;
            item.ProducedAt = DateTime.UtcNow;
            item.ResultJson = $"Updated: {DateTime.Now:HH:mm:ss.fff}";
        }
    }
    catch (Exception ex)
    {
        _logger.LogError(ex, "Error updating indicator value for {Key}", key);
    }
}
```

```
ReusableObserver<IReusable> //great for most result types
ReusableObserver<ISeries> //legal for the rest, but useless without unboxing
ReusableObserver<DonchianResult> //strongly-typed, multi-value analysis
ReusableObserver<IchimokuResult> //strongly-typed, multi-value analysis
```

All for this:
`item.Value = result?.Value ?? 0.0;`

DaveSkender:
> now I can repeatedly poll all my hubs and copy Value somewhere ... yuk

I'm working up some better examples to demonstrate how to use Hubs in an event-driven streaming architecture.

- #1899 

In a nutshell, if you're polling hubs ... you're doing it wrong! :neckbeard:

→ you really do need a WebSocket or SSE implementation for quote feeds,
or at least fake it by only polling your market data provider to simulate one

```plaintext
WSS/SSE 
   → QuoteHub 
      → [many] IndicatorHubs 
         → Reactive process [subscribes to needed IndicatorHubs]
            → [condition met]
               → [do something (e.g. make/suggest trade)]
```

The referenced PR will also include a `StrategyHub` which I'm hoping will make it easier to implement the "Reactive process" aspect of this -- as of right now this `StrategyHub` I'm drafting (#1899) is not close to being usable.

DaveSkender:
As a utility class, I think if anything, we might consider something like:

```csharp
public abstract class HubObserver<TIn> : IStreamObserver<TIn>
    where TIn : ISeries
{
    …
    // all the implementations, some as `virtual`
    // from src/_common/StreamHub/StreamHub.Observer.cs

    // e.g. has clean standard stuff like Subscribe()
}
```

then users can do whatever they want in implementations outside the library, like `ReusableObserver()`

```csharp
public class MyObserver<T> : HubObserver<T>
    where T : IReusable
{
    // whatevers or overrides
}
```

```csharp
// usage
QuoteHub quoteHub = new();
SmaHub smaHub = quoteHub.ToSma(); // subscribes to quotes
MyObserver myObserver = new();
smaHub.Subscribe(myObserver):  // subscribes to SMA (long form)

HubCollection<IReusable> = [
    quoteHub,
    smaHub,
    myObserver,
    …
];
```

DaveSkender:
> doesn't work due to invariance in the interfaces

- I fixed some of this in #1900

JGronholz:
I've realized what I was trying to implement was just a SomethingChanged notification:

```
var observer = new ReusableObserver<IQuote>(
    isSubscribed: () => true,
    onError: error =>
    {
        _logger.LogError(error, "Error in QuoteHub subscription for {Symbol} {Interval}", _selectedSymbol, _selectedTimeframe);
    },
    onCompleted: () =>
    {
        _logger.LogInformation("Hub subscription completed for {Symbol} {Interval}", _selectedSymbol, _selectedTimeframe);
    },
    rebuild: () =>
    {
        if (_isActive) RequestRefresh();
    },
    onRebuild: (fromTimestamp) =>
    {
        if (_isActive) RequestRefresh();
    },
    rebuildIndex: (fromIndex) =>
    {
        if (_isActive) RequestRefresh();
    },
    rebuildTimestamp: (fromTimestamp) =>
    {
        if (_isActive) RequestRefresh();
    },
    onAdd: (item, notify, indexHint) =>
    {
        if (notify)
        {
            if (_isActive) RequestRefresh();
        }
    });
```

The indicators update when you Add() to the quotehub...
Why do I need the hub to tell me that I added a candle to the hub?!
I don't know, either.

The upgrades this provoked are appreciated, but my original use-case was/is dumb. 👍

DaveSkender:
**_"I have not failed. I've just found 10,000 ways that won't work."_**
_\- Thomas Edison re: progress on inventing light bulb_

It's all helpful!  Appreciated, for sure.

PR #1904, opened 2026-01-12
feat: Add quote stream aggregator

JGronholz:
When a stand-alone hub is subscribed to another provider, it remains stand-alone.
Stand-alone caches do not Rebuild(), which breaks the downstream chain.
Wrong:
```
QuoteHub provider_5m = new(_maxCacheSize);
QuoteAggregatorHub aggregator = new(provider_5m);
QuoteHub provider_HigherTF = new(_maxCacheSize);
provider_HigherTF.Subscribe(aggregator); //still stand-alone
```

Correct:
```
QuoteHub provider_5m = new(_maxCacheSize); //stand-alone provider
QuoteAggregatorHub aggregator = new(provider_5m); //created with provider
QuoteHub provider_HigherTF = new(aggregator); //created with provider
```

Merging this PR will cause stand-alone providers to throw an exception if you try to Subscribe() them to another provider.
Alternatives involved extensive changes.

After merge:
```
QuoteHub provider_HigherTF = new(_maxCacheSize); //stand-alone
provider_HigherTF.Subscribe(aggregator); //throws InvalidOperationException
```


JGronholz:
There were a few subtle issues with this that required multiple passes to fully solve.
Mainly:
1) Aggregators needed to round timestamps down to their interval
5-minute bars (1:00, 1:05, and 1:10) are the same 15-minute bar (1:00)

2) The base StreamHub.Rebuild() method has duplicate detection logic (via LastItem) that prevents re-adding bars with the same timestamp during a rebuild.
For aggregators, this causes issues because:
-The cache is cleared during rebuild
-When attempting to re-add bars, they're detected as duplicates and skipped
-The cache ends up empty after rebuild
The solution was to implement custom rebuild logic that directly populates the cache without going through the base class's AppendCache method.

JGronholz:
The problem had two parts:

**Aggregator level:** During rebuild, calling OnAdd() triggered duplicate detection when multiple provider items rounded to the same aggregated timestamp.
This applied to both aggregators.
-Every quote added to a [TF] aggregator needed the timestamp rounded down to [TF].
-Every tick added to an aggregator needed the timestamp rounded down to the current/previous bar boundary.
-This creates "duplicates" that must be handled precisely
Without an upstream provider, QuoteHub rebuilds its' own cache.
The aggregators needed to do approximately the same.
Before this fix, they were rebuilding their cache from an empty cache.
It took copilot about 15 minutes to hammer each one out. Fascinating to watch...

**Provider level:** TickHub's base AppendCache logic triggered rebuilds for same-timestamp items, clearing earlier ticks from cache.
TickHub was not managing duplicates like QuoteHub and needed similar considerations.
This was a nuance with hubs with an upstream provider (stand-alone hubs were ok).

src/_common/Quotes/Quote.AggregatorHub.cs
```diff
@@ -205,6 +229,124 @@ private static Quote CreateOrUpdateBar(Quote? existingBar, DateTime barTimestamp
...
+        DateTime aggregatedTimestamp = timestamp.RoundDown(AggregationPeriod);
+
+        // Find the last bar before the rollback timestamp
+        // We need to search backwards from the cache
+        Quote? lastBarBeforeRollback = null;
+        for (int i = Cache.Count - 1; i >= 0; i--)
+        {
+            if (Cache[i].Timestamp < aggregatedTimestamp)
+            {
+                lastBarBeforeRollback = new Quote(
```
DaveSkender: This looks like an unnecessary recast.

src/_common/StreamHub/StreamHub.Observable.cs
```diff
@@ -27,7 +27,7 @@ public bool HasSubscriber(IStreamObserver<TOut> observer)
         => _observers.Contains(observer);
 
     /// <inheritdoc/>
-    public IDisposable Subscribe(IStreamObserver<TOut> observer)
+    public virtual IDisposable Subscribe(IStreamObserver<TOut> observer)
```
DaveSkender: Rather than making this overridable, which opens a different safety problem, this could throw conditionally instead

```csharp
if(observer.Provider is BaseProvider) throw;
```

PR #1910, opened 2026-01-15
fix: StochRsiHub provider notification order

JGronholz:
Initial problem:
IndexOutOfRangeException at `rsiHub.Cache[i];`

Investigating led me here:
provider > rsiHub
provider > stochRsiHub
Q: When provider triggers rebuild, who goes first?
A: RsiHub needs to always go first.

So, we just make rsiHub the stochHub's provider:
provider > rsiHub > stochRsiHub

Changes:
- Constructor chains through RsiHub (base(provider.ToRsiHub()) not base(provider))
- ToIndicator casts item to RsiResult and uses it directly
- RollbackState replays from ProviderCache (RsiResults) without triggering rsiHub rebuild
- Eliminates infinite recursion and synchronization issues
- Results match Series implementation exactly

### DIFF
--- a/src/s-z/StochRsi/StochRsi.StreamHub.cs
+++ b/src/s-z/StochRsi/StochRsi.StreamHub.cs
@@ -7,11 +7,6 @@ public sealed class StochRsiHub
     : ChainHub<IReusable, StochRsiResult>
 {
     /// <summary>
-    /// Internal RSI hub for incremental RSI calculation
-    /// </summary>
-    private readonly RsiHub rsiHub;
-
-    /// <summary>
     /// Rolling windows for O(1) RSI max/min tracking
     /// </summary>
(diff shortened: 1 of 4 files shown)


src/s-z/StochRsi/StochRsi.StreamHub.cs
DaveSkender, line 25: Oh, I see.  Yah, the use of an internal hub did smell bad.  The fix here is to force it into a normal chain like we have for GatorHub—ish.  Though I see some improvements for GatorHub too.

Ultimately it needs to behave like a proper chain.

When `quotes.ToStochRsi(lookbackPeriods)` happens it needs to swap provider for proper chain behavior.  internal provider becomes:

```csharp
var provider = quotes.ToAtrHub(lookbackPeriods);
```

PR #1944, opened 2026-01-28
fix: Render Renko chart correctly

DaveSkender:
### Motivation
- Renko bricks produce a non-linear timeline with repeated timestamps and fewer result rows than source quotes, which the chart component did not handle correctly. 
- The docs needed an embedded Renko example that signals the chart renderer to use non-linear time mapping so the visual aligns with Renko bricks rather than assuming linear, unique dates.

### Description
- Add optional `metadata.timeScale` and `chartType` handling to the chart payload and map `candles`/non-linear data to overlay rendering. 
- Implement `usesNonLinearTime` and `createTimeContext` in `docs/.vitepress/components/IndicatorChart.vue` to detect repeated/non-linear timestamps and resolve series point positions to candle indices when necessary. 
- Propagate a `timeContext` to `setupCandlestickSeries`, `setupVolumeSeries`, and `setupIndicatorSeries` so all candlestick, volume, indicator points, thresholds, and markers align with the Renko candle indices instead of raw ISO dates. 
- Embed the Renko example in `docs/indicators/Renko.md` and mark `docs/.vitepress/public/data/Renko.json` with `

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6979a0d788608326887d6a430085da44)

### DIFF
--- a/docs/.vitepress/components/IndicatorChart.vue
+++ b/docs/.vitepress/components/IndicatorChart.vue
@@ -52,7 +52,8 @@ interface ChartData {
     symbol?: string
     timeframe?: string
     indicator?: string
-    chartType?: 'overlay' | 'oscillator'
+    chartType?: 'overlay' | 'oscillator' | 'candles'
+    timeScale?: 'linear' | 'nonLinear'
     thresholds?: ThresholdLine[]
   }
   candles: Array<{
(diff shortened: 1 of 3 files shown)


DaveSkender:
@coderabbitai full review

PR #2052, opened 2026-05-30
docs(stream): Harden streaming-guide examples and fix offline SSE emulator

DaveSkender:
## Summary

Phase 1 of the pre-stable streaming hardening backlog (§E of `docs/plans/streaming-indicators.plan.md`): the zero-risk documentation + tooling fixes. **No engine code changes** — every fix is a documentation correction or the one-line offline-emulator repair. Targets `v3` (the §E plan section landed in `v3` via #2051); this is the base of the Phase 1→3 stacked series.

Implements **SR011, SR012, SR013, SR014, SR015, SR016, SR021** plus the v3.0 documentation-half of **SR004, SR005, SR006** (their v3.1 code-halves stay tracked as SR004c / SR005c / SR006c).

## What changed

**Non-compiling examples (High):**

- `stream.md` rollback example called `quoteHub.Remove(badQuote)` — no such method exists on the streaming surface. Now uses `RemoveRange(badQuote.Timestamp, notify: true)` (and re-`Add` for in-place correction), and the prose names the real removal API. *(SR011)*
- `migration.md` flagship v2→v3 StreamHub example used `ToSma` / `ToRsi` / `ToMacd` — series methods that don't resolve on a `QuoteHub` receiver. Now `ToSmaHub` / `ToRsiHub` / `ToMacdHub`. *(SR012)*

**Doc accuracy:**

- Empty-collection `Results[^1]` / `smaList[^1]` throws `ArgumentOutOfRangeException`, not `IndexOutOfRangeException` — corrected in `stream.md`, `migration.md`, `buffer.md`. *(SR016)*
- The "minimize cache size" tip rewritten to warn about cache-size inheritance and the warmup floor (RSI ~2×, TEMA/TRIX ~3×, Hilbert-transform trendline fixed 63), which otherwise throws `ArgumentOutOfRangeException` at hub construction. *(SR013)*

**Contract clarity:**

- Thread-safety section now states the **single-writer** model explicitly and names the mutating methods that must be serialized. The WebSocket example funnels `Add` through a single-writer lock instead of calling it unsynchronized from a concurrent async handler. *(SR014)*
- New **"Fault handling and recovery"** section documents `IsFaulted` / `ResetFault()` / `OverflowException` and the 100-identical-repeat threshold, plus a `Reinitialize` single-writer caveat. *(SR015, SR006 doc-half)*
- `custom-observers.md` documents the **no-throw observer contract** — a throwing callback aborts notification to sibling observers, and the same hazard applies to `OnError`. *(SR004 + SR005 doc-half)*

**Tooling (Medium):**

- `tools/sse-server/Program.cs` constructs `RandomGbm(bars: 0)` **per request** inside the `/quotes/random` handler, and `RandomGbm`'s ctor threw `ArgumentException` on `bars <= 0`, so the endpoint returned **HTTP 500** — the documented offline streaming-validation path delivered zero quotes. Fixed at the root: `RandomGbm` now accepts `bars: 0` as a legitimate empty generator (`bars <= 0` → `bars < 0`), and the server keeps the clean `bars: 0`, filling incrementally via `generator.Add(...)`. *(SR021)*

## Verification

- **Emulator fixed end-to-end:** ran the built server and hit `/quotes/random?batchSize=3` → **HTTP 200** with 3 well-formed `event: quote` SSE frames and monotonically increasing timestamps (was HTTP 500 before the fix).
- **Every changed C# snippet verified to compile against the shipped API** (`RemoveRange` / `RemoveAt`, `ToSmaHub` / `ToRsiHub` / `ToMacdHub`, `IsFaulted` / `ResetFault` / `OverflowException`), and every behavioral claim cross-checked against `src/_common/StreamHub/`.
- `npx markdownlint-cli2` clean across all changed docs.
- Local CodeRabbit-style review pass before opening: no defects.

## Scope / sequencing

First of a stacked series implementing §E. Next up: **Phase 2** (SR002 same-timestamp correction fix + SR018 test; SR008 BufferList out-of-order contract), then **Phase 3** (decision-gated SR003 / SR001). Plan checkboxes for the items above are ticked in `docs/plans/streaming-indicators.plan.md` within this PR.


docs/guide/stream.md
```diff
@@ -108,11 +108,14 @@ quoteHub.Add(quote2);
...
-quoteHub.Remove(badQuote);   // triggers recalculation
+// correct a bad value: re-add a quote with the same timestamp
+quoteHub.Add(correctedQuote);  // same timestamp, revised values → rebuild
+
+// or remove a bad quote (and everything at/after its timestamp)
+quoteHub.RemoveRange(badQuote.Timestamp, notify: true);  // triggers recalculation
 ```
 
-The hub automatically handles state rollback and recalculation when data arrives out of order or needs correction.
+The hub automatically handles state rollback and recalculation when data arrives out of order or needs correction. To revise a single value in place, re-`Add` a quote with the same `Timestamp`; to drop quotes, use `RemoveRange(fromTimestamp, notify)` or `RemoveAt(cacheIndex)`. There is no `Remove(quote)` method on the streaming surface.
```
DaveSkender: Should there be a `Remove(quote)` on the streaming surface or is that majorly problematic? I know it's not always possibly usable depending on the mode its in for chaining; there may have been a good reason for omitting this?

docs/guide/stream.md
```diff
@@ -123,12 +126,12 @@ The hub automatically handles state rollback and recalculation when data arrives
 
 ## Thread safety
```
DaveSkender, line 127: We've tried to limit the interface so that users can't do bad things directly to internal caches directly (e.g. manipulating with native `List` or `Collection` methods); however, we may not have completely plugged all holes that may lead them to problematic destructive behavior that the buffer list and stream hubs can't heal from automatically.

Since most users may not be experts with understanding thread safety, it might be helpful to assume a lower base user knowledge and have examples of what is okay and what to avoid. It's also possibly worth noting that many native .NET methods are not thread safe and that this is not unique to our implementation.

docs/guide/stream.md
```diff
@@ -218,12 +221,17 @@ void ProcessLiveData(Quote quote)
 
 This example demonstrates how to connect stream hubs to a live WebSocket feed. The pattern applies to any real-time data source (WebSocket, SSE, message queue, etc.) where quotes arrive asynchronously. The hub's `Add` method integrates each incoming quote, automatically propagating updates to all subscribed indicators.
 
+Because socket callbacks can fire concurrently, every `Add` must go through a single writer (see [Thread safety](#thread-safety) above). The handler below serializes with a lock; for higher throughput, post incoming quotes onto a single-consumer `Channel<Quote>` and call `Add` from one drain loop instead.
```
DaveSkender, line 235: We might want to clarify that users should only be adding to base hubs, not directly to subscribed ones. As a side issue, can we prevent or make inert attempts on certain problematic public interfaces by checking the kind of hub (base or subscribed, etc.).

tools/sse-server/Program.cs
```diff
@@ -59,9 +59,12 @@
...
         $"[Random] Starting stream - delivery: {interval}ms, quoteInterval: {quoteInterval}, batchSize: {batchSize?.ToString(CultureInfo.InvariantCulture) ?? "unlimited"}");
 
-    // Use Test.Data.RandomGbm for random quote generation
+    // Use Test.Data.RandomGbm for random quote generation.
+    // RandomGbm requires bars >= 1 (it rejects bars <= 0), so seed it with a
+    // single warm-up bar that is never streamed: the loop below only emits the
+    // most recently appended quote (generator[^1]) on each iteration.
     PeriodSize periodSize = Test.SseServer.Utilities.ParseQuoteIntervalToPeriodSize(quoteInterval);
-    RandomGbm generator = new(bars: 0, seed: 1000.0, periodSize: periodSize);
+    RandomGbm generator = new(bars: 1, seed: 1000.0, periodSize: periodSize);
```
DaveSkender: This seems like a hack and am not sure I've seen a problem with the prior implementation. Confirm this is a correct fix and not an unnecessary hack.

DaveSkender (commented):
Some things to consider

docs/guide/stream.md
DaveSkender: Good question — I dug into this. There's no `Remove(IQuote)` today because removal is expressed by **position** (`RemoveAt(cacheIndex)`) or **timestamp** (`RemoveRange(fromTimestamp, notify)`), and a value-based `Remove(quote)` carries two footguns: (1) match ambiguity (by timestamp only? full OHLCV equality? what if not found / multiple match), and (2) like `Add`, it's only coherent on the **root** hub — calling it on a subscribed/chained hub would desync it from its provider. So the omission reads as deliberate rather than problematic.

A thin convenience `Remove(IQuote)` on the root `QuoteHub`/`TickHub` (find-by-timestamp → `RemoveAt`/`RemoveRange`; no-op-or-throw if absent) is feasible if you want the ergonomics — happy to spec it as a v3.1 item or implement it now, just say the word. For this PR I reworded the doc to frame removal positively (by position/timestamp) and steer mutations to the root hub, rather than dwelling on the absence (`09129462`).

docs/guide/stream.md
DaveSkender: Agreed — addressed in `09129462`. Added a **"What's safe, and what to avoid"** subsection that (a) assumes lower base knowledge with ✅/❌ do-and-don't bullets, (b) notes the library intentionally limits the mutation surface (`Results` is read-only; you can't reach the cache via `List`/`ICollection` methods) but that mutating the *wrong* way can still cause damage the hub can't auto-heal, and (c) explicitly says the single-writer expectation isn't unique to us — `List<T>` / `Dictionary<TKey,TValue>` / `Queue<T>` are likewise unsafe for concurrent writers.

On "not completely plugged all holes": the main residual hole is exactly the subscribed-hub mutation case — see my reply on the WebSocket comment re: making those inert.

docs/guide/stream.md
DaveSkender: Done on the docs side (`09129462`): the new thread-safety subsection and the rollback paragraph both now say to apply `Add`/removals to the **root** `QuoteHub`/`TickHub`, never to a subscribed/chained hub (which would desync it from its provider).

On the code side ("make inert attempts on subscribed hubs"): the hub already knows whether it's self-rooted (`QuoteHub._isStandalone`; non-root hubs hold a real `Provider` rather than the inert `BaseProvider<T>`). We *could* make `Add`/`RemoveAt`/`RemoveRange` throw `InvalidOperationException` (or no-op) on a non-root hub by checking that. It's a real, contained guard but a public-behavior change, so I'd rather land it as its own decision than fold it into this doc PR — want me to spec it as a v3.1 item (or implement now)? It pairs naturally with the `Remove(IQuote)` question on the other thread.

tools/sse-server/Program.cs
DaveSkender: Confirmed the bug was real, and shipped a cleaner fix (`09129462`).

The original `RandomGbm(bars: 0)` is constructed **inside** the `/quotes/random` request handler, and `RandomGbm`'s ctor threw `ArgumentException: "Number of bars must be greater than zero."` on `bars <= 0` — so the endpoint returned **HTTP 500** on every request. It only fails when that specific endpoint is hit (lazy per-request construction), which is likely why it went unnoticed. I reproduced it just now: `bars: 0` → `HTTP 500` with that exact exception in the server log.

The `bars: 1` warm-up bar was indeed a hack. Replaced it properly: `RandomGbm` now accepts `bars: 0` as a legitimate empty generator (no test depended on it throwing), and the server is back to the clean `bars: 0`, filling incrementally — verified **HTTP 200** with quotes, and the seed now starts cleanly at 1000 instead of drifting through a discarded bar.

docs/guide/stream.md
DaveSkender: Yes, spec it as v3.1

DaveSkender: Done — spec'd as **ARCH-V31-12** in `docs/plans/streaming-indicators.plan.md` (commit `f041aa14`):

- Guard the mutating surface (`Add` / `RemoveAt` / `RemoveRange` / `Reinitialize`) on subscribed (non-root) hubs via the existing self-rooted check (`_isStandalone`; non-root hubs hold a real `Provider`, not the inert `BaseProvider<T>`) — throw `InvalidOperationException` or a documented no-op, so a leaf hub can't be desynced from its provider. Public-behavior change → lands with tests, and `stream.md` gets tightened from *documenting* the rule to *enforcing* it.
- Folded in the companion `Remove(IQuote)` root-hub convenience (from the SR011 thread) as a "ship together?" decision, since both share the root-only constraint.

(Numbered 12 because **ARCH-V31-11** is the SR003 prune-stable item on the sibling #2055 branch; the gap closes when the stack merges.)

Separately addressed the CodeRabbit nitpick on this PR: added `RandomGbmAllowsZeroBarsThenAddProducesData` (`TestData.Tests.cs`) pinning the new `bars: 0` → empty-generator path (and that `bars < 0` still throws).

PR #2062, opened 2026-05-31
fix(stream): Synchronize the observer set and close the reinitialize gap

DaveSkender:
## Summary

Closes two small but real correctness races in the streaming engine. Fourth of the stacked §L "v3.0 stable-readiness" PRs.

Implements **SR007** and **SR006c** from `docs/plans/streaming-indicators.plan.md` (§L).

> **Stacked on #2061** (→ #2060 → #2059 → v3). Base is `feat/stream-snapshot`; review only the latest commit.

## SR007 — observer-set race

`_observers` (a `HashSet`) was mutated by `Subscribe`/`Unsubscribe`/`EndTransmission` with no synchronization against the notify fan-out, which snapshots the same set. Constructing a downstream hub concurrently with a provider `Add` could throw (collection modified during enumeration) or corrupt the set.

- All `_observers` access now goes through a dedicated `_observersLock`. A `SnapshotObservers()` helper copies under the lock; **callbacks run outside the lock** so a callback can re-enter `Subscribe`/`Unsubscribe` without deadlocking.
- `ObserverCount`/`HasObservers`/`HasSubscriber` and `EndTransmission`'s clear are synchronized; `DisposeChain`'s walk now uses `SnapshotObservers()` too (the one site the SR017 PR missed).
- The subscription token is simplified to a single idempotent `Action` (`Interlocked.Exchange`).
- Lock hierarchy is strictly **`CacheLock` → `_observersLock`**, never reversed → no deadlock.

## SR006c — reinitialize rebuild→subscribe gap

`Reinitialize()` rebuilt the cache, then subscribed; an item the provider appended in that window was neither replayed nor delivered. Now it captures `ProviderCache.Count` before `Subscribe` and runs a single catch-up `Rebuild()` only if the provider grew — the quiescent case skips it. (Reinitialize is construction-only after the mutation-guard PR.) The stale race-condition TODO is removed.

## Tests

New `RaceGuards` suite, each **verified by sabotage** (fail-before / pass-after):
- `Reinitialize_WhenProviderGrowsDuringSubscribe_FoldsInGapItem` — deterministic via a provider double that injects an item during `Subscribe`; fails (30≠31) with the catch-up disabled.
- `Subscribe_ConcurrentWithNotify_DoesNotCorruptObserverSet` and `EndTransmission_ConcurrentWithNotify_DoesNotThrow` — single-writer feed vs. concurrent subscribe/teardown over a 64+ observer set; both fail when the reader-side lock is removed.

Full local run green: **indicators 2441 / 12 baseline-skipped**, **integration 5**, **public-api 71** (net10.0). Strict warnings-as-errors build clean; format + markdownlint clean.

## Self-review note

A self-review confirmed the lock ordering is deadlock-free and the SR006c double-rebuild is idempotent, and flagged the one remaining unsynchronized `_observers.ToArray()` in `DisposeChain` (from the SR017 PR) — folded into this PR for consistency.


DaveSkender:
Closing in favor of a clean, IDisposable-free rebuild.

This branch was stacked on the StreamHub `IDisposable` / `DisposeChain` work that has been **deferred from v3.0 → #2068**, and its feature commits are interleaved with that code (e.g. `ThrowIfDisposed` calls, edits to the same `StreamHub.Observable.cs`). Rebasing it cleanly is really a re-authoring, so per the maintainer steer we're rebuilding the underlying fix fresh on `v3` in a new PR — simpler solution, one at a time.

Underlying problems to re-solve: **SR007 — synchronize the _observers set against the notify fan-out**, and **SR006c — close the Reinitialize rebuild→subscribe gap.**

Branch kept temporarily for reference / lessons learned; will be deleted once the fresh PR lands.

PR #2063, opened 2026-05-31
test(stream): Add StreamHub concurrency safety-net suite

DaveSkender:
## Summary

Adds an end-to-end concurrency safety-net suite (through real indicator hubs) validating the documented single-writer / thread-safety contract. Fifth of the stacked §L "v3.0 stable-readiness" PRs — **test-only**.

Implements **TC-V31-1** from `docs/plans/streaming-indicators.plan.md` (§L).

> **Stacked on #2062** (→ #2061 → #2060 → #2059 → v3). Base is `feat/stream-observer-races`.

## Tests (`StreamHub.Concurrency.Tests.cs`)

- **Out-of-order parallel producers converge to Series** — N producers feed strided subsets through a single-writer gate; threads interleave, so quotes arrive globally out of order and drive rollback/replay. A deep `Quote→Ema→Sma` chain converges bit-for-bit (`.IsExactly`) to the batch Series. The global-minimum quote is seeded first so no out-of-order arrival is "before head" (a standalone QuoteHub drops before-head quotes by design — SR001, a pending maintainer decision left untouched).
- **Concurrent `Snapshot()` readers during a feed** — reader threads never throw and every snapshot is internally consistent (sorted); the final results match the Series.
- **Subscribe/dispose observer churn during a feed** — a stable observer stays correct and nothing faults.
- **Gated parallel feed + late arrival + removal** converges to the revised Series.

Tests #2 and #3 use a `ManualResetEventSlim` rendezvous so the reader/churn threads are guaranteed to overlap the feed (rather than racing the thread-pool spin-up).

Verified stable across repeated runs. Full local run green: **indicators 2445 / 12 baseline-skipped** (net10.0); format clean.

## Self-review note

A self-review confirmed the convergence oracles (EMA→SMA chain, revised-series-after-removal) are correct and the seed-the-minimum trick closes the before-head window; it flagged that tests #2/#3 could finish feeding before the worker threads spun up (verifying little) — addressed with the rendezvous barrier.

### DIFF
--- a/docs/plans/streaming-indicators.plan.md
+++ b/docs/plans/streaming-indicators.plan.md
@@ -168,7 +168,7 @@ The v3 streaming engine is the headline of this release after a long development
 
 #### Test coverage — prove the rollback/concurrency engine before the tag
 
-- [ ] **TC-V31-1 — StreamHub concurrency test suite** (4–6 hours). Parallel `Add` from N threads, interleaved late arrivals, concurrent observer subscribe/dispose. The safety net validating the documented single-writer / thread-safety contract.
+- [x] **TC-V31-1 — StreamHub concurrency test suite** (4–6 hours). Parallel `Add` from N threads, interleaved late arrivals, concurrent observer subscribe/dispose. The safety net validating the documented single-writer / thread-safety contract. *(PR #2063)*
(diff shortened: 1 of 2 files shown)


DaveSkender:
Closing in favor of a clean, IDisposable-free rebuild.

This branch was stacked on the StreamHub `IDisposable` / `DisposeChain` work that has been **deferred from v3.0 → #2068**, and its feature commits are interleaved with that code (e.g. `ThrowIfDisposed` calls, edits to the same `StreamHub.Observable.cs`). Rebasing it cleanly is really a re-authoring, so per the maintainer steer we're rebuilding the underlying fix fresh on `v3` in a new PR — simpler solution, one at a time.

Underlying problem to re-solve: **TC-V31-1 — StreamHub concurrency safety-net suite.**

Branch kept temporarily for reference / lessons learned; will be deleted once the fresh PR lands.

Issue #2081, opened 2026-06-10
BufferList opt-in strict mode: throw on out-of-order input


DaveSkender:
## Problem

`BufferList<T>` is a single-pass, append-only accumulator: out-of-order, duplicate, or revised timestamps produce **silently incorrect results**. This is well-documented (class remarks, buffer guide, ordering-contract characterization tests), but there is no opt-in way to detect it at runtime — bugs slip through when test data happens to be sorted and production data is not.

## Proposal

An opt-in strict mode that converts silent corruption into a fast failure:

```csharp
SmaList sma = new(20) { ThrowOnOutOfOrder = true };
sma.Add(olderQuote); // throws InvalidOperationException
```

- Cost when enabled: one timestamp comparison against the last appended item per `Add`
- Default remains off (zero overhead, current behavior, existing ordering-contract tests unchanged)
- Message should point users to StreamHub for feeds that legitimately deliver out-of-order data

*From the v3 use-case/ergonomics assessment (severity: Low, but cheap to implement). See PR #2073 for context.*

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2076, opened 2026-06-10
Warm-start state capture and restore for StreamHub and BufferList


DaveSkender:
## Problem

Restarting a 24/7 streaming service currently requires replaying full quote history to re-converge stateful indicators — EMA-family indicators need 250+ preceding bars for documented precision, and there is no way to persist or rehydrate hub/buffer-list state across process restarts. Every production deployment reinvents a persistence layer (save last N results + replay enough history to converge).

## Proposal

A state-capture API for warm starts:

- `hub.CaptureState()` → serializable snapshot of the result cache plus internal indicator state
- `hub.RestoreState(snapshot)` → rehydrate without replaying source history
- Same pattern for `BufferList<T>` implementations

The internal `RollbackRing<TState>` snapshot pattern introduced in #2073 is a natural foundation: each retrofitted hub already knows how to capture and restore its exact scalar state bit-exactly. This would generalize that capability into a (de)serialization contract.

## Considerations

- State must remain bit-exact on restore (Series parity contract)
- Versioning of serialized state across library upgrades (reject or migrate stale snapshots)
- The `Reinitialize()` rebuild→subscribe gap also matters here; a restored hub must resync cleanly with its provider

*From the v3 use-case/ergonomics assessment (severity: High). See PR #2073 for context.*

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1816, opened 2025-12-30
fix: Update performance baselines

DaveSkender:
This pull request introduces several improvements to the performance testing infrastructure and documentation, as well as critical performance fixes for streaming indicators. The most important changes include enhancements to the GitHub Actions workflows for performance testing, a comprehensive update to the streaming indicators implementation plan documenting recent performance optimizations, and code/documentation updates reflecting these improvements.

- fixes GitHub Actions performance workflows
- updates performance baselines reports
- fixes a couple hot spot indicators (Force Index)
- updates tasks for overall plan to complete performance tuning work

PR #2087, opened 2026-06-14
perf: Refresh core baselines (RG001) + fix perf workflow + reconcile streaming plan

DaveSkender:
> [!CAUTION]
> Do a soft-rebase to restart this since it was based on old v3 branch with hundreds of commits not in `main`
 
Closes the core of **RG001** (refresh performance baselines) and folds in the related loose ends discovered along the way.

### 1. Refreshed baselines — Series / Buffer / Stream

Regenerated `Performance.{Series,Buffer,Stream}Indicators-report-full.json`, replacing numbers last refreshed **2025-12-30 (#1816)**. The old set was stale on multiple axes (confirmed by diffing old vs new):

- predate the #2073 memory/GC + per-tick hot-path rewrites;
- Series benchmarks renamed `ToXxx` → `ToXxxBatch` (zero method overlap);
- `BetaHub`/`CorrelationHub`/`PrsHub` removed (non-streamable, Series-only pending `JoinHub`), 83 → 80.

**Generated on the canonical reference host** — the same machine the prior baselines came from (13th-gen i9-13900H · Windows 11 build 26200 · BenchmarkDotNet 0.15.8 · .NET 10.0.301) via `dotnet run -c Release -- --job short --filter '…Series…' '…Buffer…' '…Stream…'`. Counts: **Series 89, Buffer 79, Stream 80**.

> Methodology: `--job short` resolves to `ShortRun(5/1/5)` vs the prior `(3/1/3)` — more iterations, tighter intervals; means stay comparable for the regression threshold.

### 2. Fixed `test-performance.yml`

The full-suite job ran the no-arg `dotnet run` → **DefaultConfig (15/15)**, which (a) doesn't match the ShortRun methodology the baselines use, and (b) post-#2082 runs **>3 h and exceeds its own `timeout-minutes: 90`** (the pre-#2082 "349 s success" was a no-op). Now passes `--job short` with an explicit class filter and raises the timeout to 360.

### 3. Documented the real methodology in `baselines/README.md`

Baselines are **local, reference-host** artifacts (not CI-generated; hardware-sensitive). Updated the "Creating a new baseline" steps to use `--job short`, state the host requirement, and fix the broken glob-into-single-file `cp`.

### 4. Reconciled the streaming plan

Marked RG001 core done; recorded the local-host methodology correction, the workflow defect, and the #2085 caveat (below). Also includes the broader post-#2073 plan reconciliation: the new `RollbackRing<TState>` primitive and its effect on SR003/ARCH-V31-11, SR009/SR010 status, `BaseProvider<T>` internalization, and cross-references to the new tracking issues #2076–#2081/#2083.

### Caveat — #2085 (merged concurrently)

Baselines were measured on v3 @ #2073, just before **#2085** (SMA `Queue<double>` buffer in the Series/Stream styles) merged. The methodology PRs that merged with it (#2089/#2091/#2092/#2093) touched only `Perf.ManualTestDirect.cs` (manual harness), so these core baselines stay methodologically valid. #2085's delta is isolated to **two rows** — Series `ToSmaBatch` and Stream `SmaHub` (~12–14% faster) — which read slightly *pessimistic*, harmless for `detect-regressions.ps1` (it only flags slowdowns). Refresh at the next full baseline run.

## Verification

- All three baseline JSONs valid, full sets, real means + allocation data, no leaked local paths.
- `test-performance.yml` validated as well-formed YAML.


DaveSkender:
- [ ] these baselines need to be set to have correct max cache size to avoid including pruning in the performance periods

PR #2100, opened 2026-06-17
docs: Add missing doc site indicator page charts

DaveSkender:
In recent PRs from facioquo/stock-charts#500 we've added missing indicators like Gator Oscillator, Pivots, Piviot Points (williams fractal), and more. So now we can have those indicator pages on our vitepress documentation site render those charts through indy-charts correctly. This pull request adds those missing charts.

@DaveSkender will follow with the manual work to add the chart image thumbnail pics, so don't try to do those here.


DaveSkender:
@claude[agent] finish implementing this pull request by adding missing charts to doc site.
Additionally, for indicator pages that only show an oscillator, ensure they include the paired (aligned) price chart above it for comparison to price action.

DaveSkender:
@claude finish implementing this pull request by adding missing charts to doc site. Additionally, for indicator pages that only show an oscillator, ensure they include the paired (aligned) price chart above it for comparison to price action.

PR #2132, opened 2026-06-30
perf: Eliminate O(n²) front-pruning in BufferList and StreamHub caches

### LINKED ISSUE #2094: perf: Improve BufferList and StreamHub pruning efficiency
## Problem

Both `BufferList<T>` and `QuoteHub` (StreamHub) exhibit **O(n²) time complexity** when processing data sets larger than their default cache/list size caps, due to the way pruning is implemented.

### BufferList<T>

`BufferList<T>` defaults `MaxListSize = 100,000`. Once the internal list reaches that cap, `AddInternal` calls `PruneList()` on **every subsequent add**:

```csharp
protected virtual void PruneList()
{
    int overflow = _internalList.Count - MaxListSize; // = 1 each time
    _internalList.RemoveRange(0, overflow);           // shifts ~100,000 elements on each call
}
```

For 500k input items, this results in ~400,000 calls to `RemoveRange(0, 1)`, each requiring ~100,000 element shifts — approximately **40 billion element moves** in total.

### QuoteHub / StreamHub

`QuoteHub` has the same pattern with `maxCacheSize = 100,000`. After the cap is exceeded, `PruneCache()` is called on every item via `IsOverflowing()`, performing `Cache.RemoveRange(0, 1)`. This also cascades to observer caches via `NotifyObserversOnPrune` / `OnPrune`, doubling (or more) the total work in a streaming pipeline.

### Observed impact

Measured with ADL indicator, 500,000 periods:

| Style | Expected | Actual (with pruning O(n²)) |
|---|---|---|
| BufferList | ~43 ms | ~7,555 ms |
| StreamHub | ~50–150 ms | ~7,645 ms |

## Suggested directions

- Replace `List<T>` backing store with a **circular/ring buffer** or `LinkedList<T>` to achieve O(1) front removal instead of O(n).
- Or **batch-prune** in larger chunks (e.g., remove 10% of `MaxListSize` at once) to amortize the `RemoveRange` cost.
- Apply the same fix to both `BufferList<T>` and `StreamHub` cache pruning, and to the `OnPrune` observer cascade.

## References

- Identified during investigation in PR #2085 (perf: Use Queue buffer consistently across SMA styles)
- Reported by @DaveSkender

DaveSkender:
Once `BufferList<T>` and `StreamHub` reach their size caps (`MaxListSize`/`MaxCacheSize`, default 100k), every subsequent add prunes one item from the front of a backing `List<T>`. Front removal shifts the entire array, so steady-state streaming degrades to **O(n²)** — measured at ~7.5s for ADL over 500k periods.

### Changes

- **New `PruningList<T>`** (`src/_common/Generics/PruningList.cs`) — `IList<T>` + `IReadOnlyList<T>` backed by a `List<T>` with a logical head offset. Front removals advance the offset in **O(1)**; the dead prefix is physically reclaimed only when it reaches the live-window size, keeping amortized removal O(1) and storage bounded to ~2x the cap. Index/count/enumeration/mutation semantics match `List<T>` over the live window, so it drops in wherever only retained (post-prune) elements are observed.
- **`BufferList<T>._internalList`** now a `PruningList<TResult>` — `PruneList`'s `RemoveRange(0, overflow)` becomes O(1) amortized; derived `PruneList` overrides are untouched.
- **`StreamHub<TIn,TOut>.Cache`** now a `PruningList<TOut>` — `PruneCache` and the `OnPrune` observer cascade become O(1) amortized.
- **Tests** for `PruningList<T>` covering live-window semantics under an active head (insert/remove/index/enumerate/`AsReadOnly`) plus a steady-state sliding-window oracle.

Exact-count semantics are preserved — caches still settle to exactly `MaxListSize`/`MaxCacheSize` — so no existing pruning tests changed.

### Impact (ADL, 500,000 periods)

| Style | Before | After |
|---|---|---|
| BufferList | ~7,555 ms | ~69 ms |
| StreamHub | ~7,645 ms | ~353 ms |

### DIFF
--- a/docs/migration/v3.md
+++ b/docs/migration/v3.md
@@ -289,32 +289,6 @@ Popular indicators with complete streaming documentation:
 - Oscillators: [RSI](/indicators/rsi), [MACD](/indicators/macd), [Stochastic](/indicators/stoch)
 - Channels: [Bollinger Bands](/indicators/bollinger-bands), [Keltner](/indicators/keltner)
 
-## Known issues and tips
-
-A few v3 behaviors are known to be sub-optimal and are tracked in our backlog. They aren't breaking — the library stays usable — but we mention them here so you can recognize the symptoms and plan around them.
-
-### Streaming performance degrades after the cache prunes
-
-Both the BufferList (`MaxListSize`) and StreamHub (`maxCacheSize`) styles keep a rolling cache that defaults to **100,000 elements**. Once that limit is exceeded, the oldest results are automatically pruned, and two symptoms follow:
-
-- **Throughput drops noticeably** once pruning begins — still functional, just slower.
-- **Early results are no longer retrievable**, and a late arrival or correction that triggers a rebuild can't see the pruned history, so rebuilt values for stateful indicators become approximate.
-
-If you replay a large historical dataset — for example, several years of minute bars — and need the full history resident or steady performance, raise the cache size so pruning never triggers:
-
-::: tip ✨ Raise the cache size for large datasets
-
-```csharp
-// StreamHub: size the root hub (it cascades to chained hubs)
-BarHub barHub = new(maxCacheSize: 250_000);
-
-// BufferList: size the list
-SmaList smaList = new(20) { MaxListSize = 250_000 };
-```
-
-`maxCacheSize` is inherited by every chained hub, so size it for the **largest warmup in the whole chain**. Each hub validates the size against its own warmup floor at construction and throws `ArgumentOutOfRangeException` if it is too small. See [Stream hub memory management](/guide/styles/stream#memory-management) for details.
-:::
-
 ## Need help?
 
 - [Guide and Pro tips](/guide/getting-started) - Getting started with v3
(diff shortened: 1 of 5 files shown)


DaveSkender:
### Independent perf validation (from PR #2133)

I re-measured the pruning path directly to judge the claims in this description. Key point: the `ManualTestDirect` harness sets `MaxListSize`/`maxCacheSize = Periods`, so it never prunes. To exercise the actual fix I added a `PERF_TEST_CAP` override (in #2133) and ran **ADL, 500,000 bars, cap = 100,000** (steady-state pruning), before vs. after this change on the same machine (i9-13900H / .NET 10):

| Style | Before (`List<T>`) | After (`PruningList<T>`) | Speedup |
|---|---:|---:|---:|
| BufferList | 6,476 ms | 40 ms | ~160x |
| StreamHub | 19,877 ms | 123 ms | ~160x |
| Series (control) | 52.6 ms | 43 ms | unchanged |

**Verdict: claims validated and, if anything, conservative.** StreamHub's real before-cost is ~20s (worse than the ~7.6s quoted) because pruning cascades through the observer chain. The Series control is unchanged, confirming the speedup is the pruning fix and not measurement noise. No code changes needed here.

PR #2133, opened 2026-06-30
perf: Refresh core baselines

DaveSkender:
Implements streaming plan task RG001 to reset performance baselines


DaveSkender:
### RG001 progress

Working the streaming-plan release gate **RG001 — refresh performance baselines**. Status:

- ✅ **Validated PR #2132's pruning fix** independently (before/after, steady-state pruning at 500k bars / 100k cap). Posted the table on #2132 — claims hold and are conservative.
- 🔧 **Harness improvement (this PR):** added a `PERF_TEST_CAP` env override to `Perf.ManualTestDirect.cs` so the steady-state pruning path can be benchmarked (cap < periods). Default = periods, preserving prior no-prune behavior.
- ⏳ **Baseline refresh in progress:** regenerating `tools/performance/baselines/*.json` on the same hardware the committed baselines were produced on (i9-13900H / Win11 / BenchmarkDotNet 0.15.8). Series + Buffer done; Stream running; Style/External/Utility classes next. The committed baselines date to ~Dec 2025 (v3.0.0, pre P004–P014 and pre-PruningList).

Next: drop in refreshed baseline reports, run a spot-check sweep for any remnant O(n²) indicators, then request review.

DaveSkender:
### RG001 complete — baselines refreshed, PR #2132 validated

**Pushed:** refreshed all 7 perf baselines + added the missing `UtilityNullMath` baseline (the `test-performance` workflow already lists it), plus a `PERF_TEST_CAP` harness override to benchmark the steady-state pruning path. Generated on the same hardware the prior baselines were (i9-13900H / .NET 10 / BDN 0.15.8). No `src` changes.

#### Verdict on the PR #2132 claims — validated, and the `PruningList` change is a strict improvement
- **Pruning path** (ADL 500k bars, 100k cap): Buffer **6,476 ms → 40 ms**, StreamHub **19,877 ms → 123 ms** (~160x). Series control unchanged. Claims hold and are conservative.
- **Common no-prune case** (502 periods): in apples-to-apples *isolated* runs, `PruningList` ≈ `List<T>` (within 0–6% noise). No common-case regression.

#### About the detector output (important context for reviewers)
Running `detect-regressions.ps1` against the **old** baselines flags Buffer ~25–45% and Stream 30–80x "regressions." These are **not** caused by #2132 and **not** real regressions in this PR — they are the exact stale-baseline artifact RG001 exists to reset:
- I reproduced both by reverting `PruningList` → `List<T>` (pre-fix code): the pre-fix `List<T>` is *also* ~100x slower than the Dec-2025 baseline on the Stream replay path. So the delta predates #2132 and matches the plan's *"pre-centralized-RollbackState"* note.
- The Stream benchmark measures subscribe-to-prepopulated-provider (full replay/rebuild per invocation); that path got heavier with the centralized `RollbackState` architecture — a known, deliberate v3 tradeoff — and is also quite noisy.

Net: the refreshed baselines encode current reality; the pruning fix is a clean win. Requesting review.

DaveSkender (commented):
Remove all copies baselines and re-run, some of these look old or partially complete in markdown

Issue #2146, opened 2026-07-06
Introduce *Extensions classes and deprecate legacy extension-holder names (non-breaking)


DaveSkender:
## Objective

Deliver the user-visible outcome of #2139 (properly named `{Area}Extensions` classes per .NET guidance) **without waiting for v4**, using the repository's established deprecate-then-remove pattern (`Obsolete.V3.*.cs`). Only the *removal* of the legacy names is breaking; that remainder stays in #2139 (v4).

## Mechanics (binary- and source-compatible)

For each public extension-holder class:

1. Add a new `{X}Extensions` class carrying the extension methods (`this` parameters).
2. Convert the legacy class's members to **non-extension** static delegates (drop the `this` modifier, keep identical signatures) marked `[Obsolete("Use {X}Extensions...")]`.
3. Migrate internal call sites to the new classes (internal builds treat warnings as errors).

Why this is safe:

- Extension-syntax call sites (`list.RemoveWarmupPeriods()`) bind to the new class — no ambiguity, because only one class carries the `this` modifier.
- Explicit static call sites (`Pruning.RemoveWarmupPeriods(list, n)`) keep compiling (with a deprecation warning) — dropping the `this` modifier only removes `ExtensionAttribute`, which does not change the method signature.
- Compiled consumers keep binding — the legacy methods still exist with identical IL signatures.

## Scope (audit each; classes in post-#2145 layout)

| Class | File | Action |
| ----- | ---- | ------ |
| `Reusable` | `src/Common/Reusable/Reusable.cs` | `ReusableExtensions` + deprecate legacy members |
| `Pruning` | `src/Common/Extensions/Generics/Pruning.cs` | `PruningExtensions` + deprecate |
| `Seeking` | `src/Common/Extensions/Generics/Seeking.cs` | `SeekingExtensions` + deprecate |
| `Sorting` | `src/Common/Extensions/Generics/Sorting.cs` | `SortingExtensions` + deprecate |
| `StringOut` | `src/Common/Extensions/Generics/StringOut.*.cs` | `StringOutExtensions` + deprecate |
| `NullMath` | `src/Common/Extensions/Math/NullMath.cs` | decide: extension holder vs BCL-style utility (utility name may stay per plan Rule 4) |
| `Candlesticks` | `src/Common/Candles/Candlesticks.cs` | decide |
| `BarIntervals` | `src/Common/Enums/BarIntervals.cs` | decide |
[...]

Issue #2139, opened 2026-07-05
Rename public extension-method holder classes to *Extensions (breaking)


DaveSkender:
## Objective

Finish plan Tasks 2.2–2.4 and 4.1: remove the deprecated legacy extension-holder class names so only the `{Area}Extensions` names remain.

**Re-scoped 2026-07-06**: the non-breaking half of this work (introduce `*Extensions` classes, convert legacy holders to `[Obsolete]` non-extension delegates, migrate internal call sites) moved to #2146 (milestone v3.1) using the repository's `Obsolete.V3` deprecate-then-remove pattern. This issue now covers only the **breaking removal** step.

## Scope (after #2146 lands)

- Delete the deprecated legacy holder classes / members: `Reusable`, `Pruning`, `Seeking`, `Sorting`, `StringOut`, plus whichever of `NullMath` / `Candlesticks` / `BarIntervals` the #2146 audit converts
- Remove the shadowed `IndicatorConfigExtensions.ToBuilder(this IndicatorConfig)` overload (permanently shadowed by the `IndicatorConfig.ToBuilder()` instance method; `[Obsolete]`-marked in #2146; flagged in PR #2141 review)
- File cleanup follows: legacy files under `src/Common/Extensions/Generics/`, `src/Common/Reusable/Reusable.cs`, etc. collapse to the `*Extensions` files only

## Rules (from the former `docs/plans/file-reorg.plan.md`, since removed; text preserved in git history)

- Extension-method classes (have `this` parameters): `{Type}Extensions` / `{Area}Extensions`
- Static utility classes (no `this` parameters): descriptive functional names following BCL patterns (like `Math`, `Convert`) — e.g. `Numerical` keeps its name
- File names follow the class names

## Why this is breaking

Deleting public static classes/members breaks compiled consumers and explicit static call sites — MAJOR version only.

## Related

- Parent: #1810
- Non-breaking deprecation phase: #2146 (v3.1)
- Plan: former `docs/plans/file-reorg.plan.md` (Tasks 2.2, 2.3, 2.4, 4.1) — file removed, content in git history

## Correction (2026-08-09)

[...]

PR #2148, opened 2026-07-06
Unify McGinley Dynamic naming as non-breaking v3.1 deprecation (#2138 Option A)

### LINKED ISSUE #2138: Unify McGinley Dynamic type family naming
## Objective

Resolve the split naming of the McGinley Dynamic indicator. It is the only indicator whose type family mixes two base names, and the base name it mixes in is a poor one for consumers.

> [!IMPORTANT]
> **Updated 2026-09-03.** Option A — standardizing on `Dynamic` — is withdrawn. `Dynamic` is a problem in consumer code, not just an internal inconsistency, so making it the family name would entrench the weakness rather than resolve it.
>
> **Milestone: v4, deferred on priority.** Both remaining options break compiled consumers, so v4 is the earliest they could land regardless. The operative reason for not pursuing it sooner is simply that it does not rank against current work — a naming inconsistency with a correct catalog display name is not worth spending a disclosed break on now. The base-name choice below (Option B or C) remains open and does not need deciding until this is picked up.

## Why `Dynamic` is the wrong base name

It is not only adjacency to the `dynamic` contextual keyword. In consumer code the type reads as dynamic typing:

```csharp
IReadOnlyList<DynamicResult> results = quotes.ToDynamic(14);
double? v = results[^1].Dynamic;     // reads as "the dynamic member"
```

The property is the sharpest instance. This library's convention is `{Indicator}Result.{Indicator}` (`EmaResult.Ema`, `SmaResult.Sma`, `ObvResult.Obv`), so the McGinley value is surfaced as `.Dynamic` — the member consumers type most often, and the one that carries no indication of what it computes. Its XML doc currently reads "Dynamic value", with no mention of McGinley.

**All three options in the original issue omitted the property rename.** Any family rename must carry it, or the most-used member keeps the defect the rename is meant to fix.

## Current state

`public static partial class MgDynamic` is opened in `MgDynamic.Series.cs`, `MgDynamic.Utilities.cs`, `MgDynamic.Catalog.cs`, and also `DynamicHub.cs` and `DynamicList.cs` — it is the per-indicator facade holding all three style entry points (see #2137). Everything else in the family is `Dynamic*`.

| Element | Name | First shipped |
| ------- | ---- | ------------- |
| Facade class | `MgDynamic` | v3 |
| `ToDynamic()` | Series entry point | v3 (v2 shipped `GetDynamic()`, shimmed) |
| `DynamicResult` + `.Dynamic` property | result record | **v2** (`sealed class DynamicResult : ResultBase`) |
| `DynamicHub`, `ToDynamicHub()` | streaming | v3 |
| `DynamicList`, `ToDynamicList()` | buffered | v3 |
| `IDynamic` | interface | v3 |
| Directory | `src/Indicators/c-d/Dynamic/` | — |

The catalog already displays the indicator correctly as `"McGinley Dynamic"` with id `"DYNAMIC"`. Only the type names disagree with it.

## Remaining options — both breaking

**Option B — `MgDynamic*`.** Rename `DynamicHub` → `MgDynamicHub`, `DynamicList` → `MgDynamicList`, `DynamicResult` → `MgDynamicResult`, `IDynamic` → `IMgDynamic`, property → `.MgDynamic`, methods `ToMgDynamic*`, directory → `MgDynamic/`. The facade already carries this name, so this reads as "the v3 streaming and buffered additions drifted from the facade; restore it." Shortest names. Cost: `Mg` is not a recognized abbreviation for McGinley.

**Option C — `McGinleyDynamic*`.** Same shape, full name, property `.McGinleyDynamic`. Matches the catalog display name exactly and is self-documenting with no abbreviation to decode. Cost: verbosity at the call site.

### Recommendation: Option C

Consumer clarity is the stated driver, and `McGinleyDynamic` is the only choice that needs no prior knowledge to read. It also makes the type names agree with the catalog display name that already ships. Option B is defensible if call-site verbosity outweighs that.

## What breaks, precisely

- Compiled consumers referencing any renamed type break at the assembly level.
- Source consumers break on the type names, on `.Dynamic`, and on explicit static calls; extension-method call sites (`quotes.ToDynamic(14)`) break only because the method itself is renamed.
- Deprecation shims do not fully cover this. The class and interface renames can carry `[Obsolete]` derived-type or dual-interface shims, but the result record cannot: a consumer typed as `IReadOnlyList<DynamicResult>` cannot receive `IReadOnlyList<McGinleyDynamicResult>`, and C# has no assembly-internal alias visible to consumers. Keeping both result records in parallel is possible but doubles the public result surface for the duration.

## Timing — settled

Deferred to v4 on priority (2026-09-03). Both options break compiled consumers, so this could not ship in a v3 minor without a deliberate, disclosed break, and it does not rank highly enough to justify one.

For reference if that is ever revisited: the v3-era pieces (hub, list, interface, and the `To*` method names) shipped with v3.0.0 on 2026-06-23, so their adoption window is comparatively short. `DynamicResult` and its `.Dynamic` property span v2 and v3 and carry the longest adoption — and are also the only part that cannot take a shim.

## Acceptance criteria (once B or C is chosen)

- [ ] Family renamed consistently: facade, hub, list, result record, interface, extension methods, directory, and files
- [ ] Result property renamed and its XML doc corrected to name McGinley Dynamic
- [ ] Catalog `MethodName` values updated; display name and `"DYNAMIC"` id confirmed unchanged
- [ ] `[Obsolete]` shims added for every case that admits one; the result record's limitation documented in the migration notes
- [ ] Migration guidance added to the release notes
- [ ] Shim removal tracked against #2139

## Work mode

🙋🤖 PAIRING (human first) — the base name and the timing are both naming judgments.

## History

Originally filed as plan Task 4.10 / appendix A.4.1 under parent #1810 (now closed). Option A (`Dynamic*`) withdrawn 2026-09-03.

DaveSkender:
## Summary

Resolves #2138 using **Option A**, delivered as a **non-breaking v3.1** change via the deprecate-then-remove pattern established in #2146 (`Obsolete.V3.*.cs`), instead of a breaking v4 rename.

- Renamed the series entry-point class `MgDynamic` → `Dynamic`, aligning it with the already-unified `DynamicHub` / `DynamicList` / `DynamicResult` / `IDynamic` family.
- Renamed source files `MgDynamic.Series.cs` / `MgDynamic.Utilities.cs` / `MgDynamic.Catalog.cs` → `Dynamic.Series.cs` / `Dynamic.Utilities.cs` / `Dynamic.Catalog.cs`, and updated all internal references (`DynamicHub.cs`, `DynamicList.cs`, `Catalog.Listings.cs`).
- Added `src/Obsolete.V3.McGinleyDynamic.cs`, keeping `MgDynamic` as a warning-level `[Obsolete]` class. Following the #2146 mechanics, its members drop the `this` modifier (no longer extension methods), so:
  - Extension-syntax call sites (`quotes.ToDynamic(14)`) bind unambiguously to the new `Dynamic` class — no ambiguity.
  - Explicit static call sites (`MgDynamic.ToDynamic(quotes, 14)`) keep compiling with a deprecation warning.
- Suppressed the expected `CA1724` analyzer warning (`Dynamic` name partially matching the `System.Dynamic` namespace) via `GlobalSuppressions.cs`, with justification referencing this deliberate naming decision.
- Removal of the legacy `MgDynamic` name remains a breaking remainder for a future major version (mirrors how #2146 splits from #2139).

## Test plan

- [x] `dotnet build` succeeds for `src/Indicators.csproj` (net9.0) and `tests/Library/Tests.Indicators.csproj` (net10.0), with the `CA1724` warning suppressed
- [x] Full `tests/Library` suite passes: 2491 passed, 12 skipped, 0 failed
- [x] `tests/PublicApi` suite passes: 76 passed, 0 failed
- [x] Manually verified extension-syntax (`Bars.ToDynamic(14)`) resolves without ambiguity and that the obsolete `MgDynamic.ToDynamic(Bars, 14)` static call still compiles (with a `CS0618` warning) and returns identical results to the new `Dynamic` class

https://claude.ai/code/session_0158G9WRd51YnKyhkvF6xfpp

---
_Generated by [Claude Code](https://claude.ai/code/session_0158G9WRd51YnKyhkvF6xfpp)_


src/GlobalSuppressions.cs
```diff
@@ -33,3 +33,10 @@
     Justification = "Temporary, during deprecation period.",
     Scope = "member",
     Target = "~P:FacioQuo.Stock.Indicators.ISeries.Date")]
+
+[assembly: SuppressMessage(
+    "Naming",
+    "CA1724:Type names should not match namespaces",
```
DaveSkender: This reminds me of why I’d renamed it `MgDynamic`. While this suppresses in our library it will create a similar nuisance alert in consumer apps.

DaveSkender (commented):
Let’s renamed it to `MgDynamic`

src/GlobalSuppressions.cs
DaveSkender: Good call — reverted the entry-point class back to `MgDynamic` (commit 890dda5c). `DynamicHub`/`DynamicList`/`DynamicResult`/`IDynamic` are untouched, the CA1724 suppression is gone since there's no longer a name collision, and the obsolete shim is removed since nothing is being renamed. Codacy now reports 0 issues / 0% coverage variation.

---
_Generated by [Claude Code](https://claude.ai/code)_

DaveSkender:
@facioquoqa review

DaveSkender:
@claude address the review feedback. If this can’t be implemented, close, update issue with comments, and cancel issue if there’s no reasonable path forward.

PR #2151, opened 2026-07-14
docs: Lean on indy-charts client-side resilience for charts

### LINKED ISSUE #2124: Simplify docs chart integration once indy-charts gains client-side resilience
## Summary

Once `@facioquo/indy-charts` gains **client-side resilience** (automatic retry/backoff + optional last-good cache — tracked in **facioquo/stock-charts#517**, P2), the docs site's chart integration can be simplified. Several pieces here exist only because the published component currently does a single `fetch` and throws on any transient failure.

> **Blocked by:** facioquo/stock-charts#517 (indy-charts P2 work). Do not start until that ships in a published `@facioquo/indy-charts` release.

## Background

PR #2123 hardened the docs charts after intermittent `No chart data is available.` outages: it removed a buggy `window.fetch` fallback that masked API blips as a sticky empty state, so transient failures now surface indy-charts' real error + **Retry** UI and self-heal on reload. That's the right behavior given today's component, but it leaves the docs site carrying compensating logic.

## Candidate simplifications (re-evaluate once #517 lands)

- [ ] **Lean on indy-charts' built-in retry** instead of relying on users manually clicking Retry. If the component auto-retries transient `5xx`/network errors, the docs charts recover from a market-open blip with no interaction.
- [ ] **Adopt a last-good / stale-render path** if indy-charts exposes one, so a refetch failure shows slightly-stale data rather than an error.
- [ ] **Simplify the dev-only fetch proxy** in `docs/.vitepress/theme/index.ts`. It currently rewrites API calls to Vite's `/chart-api-proxy` to dodge CORS in local dev. Keep whatever is still needed for CORS, but drop anything that overlaps with new component behavior.
- [ ] **Reduce test scaffolding** in `docs/tests/charts.spec.ts` + `docs/.vitepress/public/data/chart-api/*.json`. If indy-charts ships a first-class static/offline data source (instead of consumers monkeypatching `fetch` and hand-maintaining fixtures), the hermetic mocks and fixtures can shrink considerably.

## Acceptance

- No regression in the three website test tasks (`Test: Website charts`, `links`, `a11y`).
- Net reduction in docs-side chart glue code, with behavior equal-or-better than PR #2123.

DaveSkender:
Closes #2124.

## Summary

`@facioquo/indy-charts@0.8.0` (facioquo/stock-charts#522) added **client-side resilience** to `createApiClient`:

- **retry** — transient failures (network / `5xx` / `429`) are retried with exponential backoff + jitter. **On by default** (`3` attempts, `500 ms` base).
- **staleCache** — each successful response is cached in `sessionStorage`; if a later refetch exhausts its retries, the last-good value is served instead of erroring. Opt-in.
- **onStale** — callback fired when stale data is served.

With that shipped, the docs charts no longer need to carry compensating logic. This PR wires the docs into the library's resilience.

## Changes

- **New `docs/.vitepress/theme/chart-api.ts`** — single source of truth for the chart-API base URL and the shared resilience options (`staleCache: true` + an `onStale` dev-console hook). Mirrors the existing `chart-theme.ts` lockstep pattern. `retry` is intentionally unset — it's on by default in 0.8.0 (and its type is `RetryConfig | false`, so a bare `true` wouldn't type-check).
- **`theme/index.ts`** — the `setupIndyChartsForVue` `api` config now spreads the shared resilience options; the duplicated hardcoded API base URL is gone; the now-outdated dev-proxy comment (describing the removed pre-#2123 fallback) is trimmed.
- **`LandingCharts.vue`** — its hand-rolled `createApiClient` gets the same base URL + resilience, so the home-page overlay self-heals too.
- The dev `window.fetch` → `/chart-api-proxy` shim **stays** — it's purely a local-dev CORS workaround (indicator endpoints are absolute API URLs), unrelated to resilience.

## Behavior

Equal-or-better than #2123: a market-open API blip now auto-retries and, after a first successful load, falls back to slightly-stale cached data rather than surfacing an error. The library's Retry/error UI remains the final fallback when all retries are exhausted with no cache.

## Testing

- `Test: Website charts (playwright)` — **84 passed** locally.
- `docs:build` succeeds (VitePress SSG).
- Links (htmlproofer) and a11y (pa11y) tasks are unaffected — changes are client-side chart-API wiring only; no markup, links, or ARIA changed.

## Notes on test scaffolding

The Playwright fixtures in `docs/.vitepress/public/data/chart-api/*.json` and the `mockStockChartsApi` route stubs are **kept**: 0.8.0's stale path is `sessionStorage`-based, not a first-class static/offline data source, so the hermetic mocks are still required for deterministic, offline tests.

### DIFF
--- a/docs/.vitepress/components/LandingCharts.vue
+++ b/docs/.vitepress/components/LandingCharts.vue
@@ -12,8 +12,8 @@ import {
 } from '@facioquo/indy-charts'
 
 import { DARK_SURFACE, LIGHT_SURFACE } from '../theme/chart-theme'
+import { CHART_API_BASE_URL, CHART_API_RESILIENCE } from '../theme/chart-api'
 
-const API_BASE = 'https://stock-charts-api.azurewebsites.net'
 const BAR_COUNT = 250
 const EMA_FAST_COLOR = '#ff4d8d'
 const EMA_SLOW_COLOR = '#26c6da'
@@ -111,7 +111,7 @@ async function renderCharts(): Promise<void> {
 
     setupIndyCharts()
 
-    const client = createApiClient({ baseUrl: API_BASE })
+    const client = createApiClient({ baseUrl: CHART_API_BASE_URL, ...CHART_API_RESILIENCE })
     const [quotes, listings] = await Promise.all([client.getQuotes(), client.getListings()])
     if (disposed || token !== loadToken) return
 
--- a/docs/.vitepress/theme/chart-api.ts
+++ b/docs/.vitepress/theme/chart-api.ts
@@ -1,0 +1,31 @@
+// Shared chart-API connection settings. Single source of truth so the VitePress
+// theme (every <StockIndicatorChart> instance) and the landing overlay talk to
+// the same endpoint with identical resilience — mirrors chart-theme.ts.
+//
+// indy-charts 0.8.0 added client-side resilience (facioquo/stock-charts#522):
+//   • retry — transient network / 5xx / 429 failures are retried with backoff.
+//     On by default (3 attempts, 500 ms base), so it is intentionally unset here.
+//   • staleCache — each successful response is cached in sessionStorage; if a
+//     later refetch exhausts its retries, the last-good value is served instead
+//     of erroring. Opt-in, enabled below.
+//
+// This lets the docs charts self-heal from a market-open API hiccup with no user
+// interaction, replacing the compensating fetch logic the docs used to carry.
+
+import type { ApiClientConfig } from '@facioquo/indy-charts'
+
+export const CHART_API_BASE_URL = 'https://stock-charts-api.azurewebsites.net'
+
+/** Note in the dev console when a chart falls back to the last-good cache. */
+function handleStale(context: string): void {
+  console.warn(`[stock-charts] Live "${context}" request failed; showing recent cached data.`)
+}
+
+/**
+ * Resilience options shared by every chart-API client. Spread into the
+ * `createApiClient` / `setupIndyChartsForVue` config alongside `baseUrl`.
+ */
+export const CHART_API_RESILIENCE: Pick<ApiClientConfig, 'staleCache' | 'onStale'> = {
+  staleCache: true,
+  onStale: handleStale
+}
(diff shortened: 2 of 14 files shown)


DaveSkender:
## Deferred: candidate simplification 4 (shrink Playwright fixtures) — recommend **won't-do**

This PR delivers candidates 1–3 (built-in retry, stale-cache fallback, dev-proxy cleanup). It **intentionally skips 4** — reducing `docs/tests/charts.spec.ts`'s `mockStockChartsApi` stubs and the `docs/.vitepress/public/data/chart-api/*.json` fixtures.

**Why it was skipped now.** 4 was written as conditional: _"If indy-charts ships a first-class static/offline data source (instead of consumers monkeypatching `fetch` and hand-maintaining fixtures)."_ `@facioquo/indy-charts@0.8.0` did **not** ship that. Its `staleCache` path is `sessionStorage`-based runtime resilience, not a test/offline data provider — so there's nothing new to lean on here.

**Recommendation: don't pursue the follow-on.** Rationale:

1. **The premise no longer holds.** The tests already intercept at the **network layer** via Playwright `page.route()` — they do *not* monkeypatch `fetch` inside the component. That's the robust, decoupled approach. An indy-charts "static data source" wouldn't remove the need to *supply* fixture data; it would just move the seam from Playwright routes to a library API. That's a lateral move, not a net reduction.
2. **Cost is cross-repo and recurring.** It would require designing + shipping a new indy-charts feature (→ 0.9.0 publish), then another docs bump + a test rewrite here — real coordination cost for marginal, possibly-negative payoff.
3. **The current scaffolding is small and stable.** One ~25-line helper + 4 JSON fixtures, hermetic, deterministic, offline, and green (84 tests). It isn't "compensating for missing retries" — it exists to keep chart tests offline/deterministic, a need that resilience features don't obviate.

**Suggested disposition:** close #2124 as completed by this PR and mark 4 as won't-do (or drop it to the backlog). If indy-charts ever grows an offline data source for its *own_ reasons (Storybook, its own test harness), we can revisit opportunistically — but I would not bump indy-charts solely to chase this.

cc #2124

PR #2173, opened 2026-08-08
fix: Address code review findings in moved utility classes

DaveSkender:
Behavioral and quality fixes for pre-existing issues surfaced by review of the file reorganization (all baseline-safe; no indicator numerical outputs change):

- **Numerical**: correct `ArgumentNullException.ThrowIfNull` calls that passed messages as paramName (pinned by `WithParameterName` assertions); add paramName to Slope length mismatch; guard Slope zero denominator per NaN policy (documented `NaN` return); replace int casts in `GetDecimalPlaces` with `decimal.Truncate` to fix `OverflowException` on volumes above `int.MaxValue`; hoist Lanczos coefficients to a static field
- **DeMath**: preserve signed zero in `Atan` and `Atan2` (matching the documented `Math.Atan2` quadrant contract); return finite result at the exact `Exp` overflow boundary (test asserts the deterministic correctly-rounded literal, not platform `Math.Exp`)
- **PruningList**: capture version and count at `GetEnumerator` time and iterate logical indexes so prefix reclamation from a no-match `RemoveAll` cannot skip elements under an active enumerator
- **BinarySettings**: validate indexer range (0–7) instead of silently wrapping shift counts
- **BarPart**: centralize `CandlePart` validation — hubs, buffer lists, and series conversions reject undefined values eagerly, including via the `BarPartList.CandlePartSelection` `init` accessor (object-initializer path); `<exception>` docs added in house style on all public entry points
- **Tests**: cover all fixes with discriminating tests (each fails on pre-fix code); assert catalog `MethodName` for stream and buffer listings; align test names with the `MethodName_StateUnderTest_ExpectedBehavior` convention; preserve operator and `Equals(object)` dispatch coverage in `BinarySettings` equality tests; rename duplicate `BarPartBufferListTests` class in utilities tests to `BarPartUtilitiesTests`

### DIFF
--- a/src/Common/Math/DeMath.cs
+++ b/src/Common/Math/DeMath.cs
@@ -156,7 +156,7 @@ internal static class DeMath
             return double.NaN;
         }
 
-        if (x >= ExpOverflow)
+        if (x > ExpOverflow)
         {
             return double.PositiveInfinity;
         }
@@ -281,7 +281,7 @@ internal static class DeMath
 
         if (x == 0d)
         {
-            return 0d;
+            return x; // preserves signed zero
         }
 
         // Range reduction: atan(x) = sign(x) * (pi/2 - atan(1/|x|)) for |x| > 1
@@ -350,7 +350,13 @@ internal static class DeMath
             return -HalfPi;
         }
 
-        // Preserve signed-zero for (0,0) by returning `y` (will be +0.0 or -0.0)
+        // x == -0.0 belongs to the negative half-plane: (±0, -0) -> ±pi
+        if (BitConverter.DoubleToInt64Bits(x) < 0L)
+        {
+            return BitConverter.DoubleToInt64Bits(y) < 0L ? -Pi : Pi;
+        }
+
+        // Preserve signed-zero for (±0, +0) by returning `y` (will be +0.0 or -0.0)
         return y;
     }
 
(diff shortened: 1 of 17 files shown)


DaveSkender:
@copilot address all unresolved comments and review feedback, including ones from other reviewers and ones from prior reviews that are still open.

<details>
<summary>Code completion checklist (must do)</summary>

Fixing the issues:

- [ ] Inspect each unresolved comment in the pull request
  - Deeply understand the context of the problem and solution suggested by the reviewer
  - Do not assume the reviewer is correct in identifying root cause or solution
  - Fix the underlying root problem
  - Reply to all open review comments, even if skipped, so I can clearly see which items you have addressed

Finishing checklist

- [ ] Run all lint auto-fix commands (ref `tasks.json`) to ensure base cleanliness of codebase
- [ ] Check for current pull request check run failures.  If present:
      - investigate root cause
      - fix all reasons failures (nothing is out of scope)
- [ ] Finally, run full swarm review and code completion checklist as defined in the repository instructions and fix all issues reported. If no instructions are provided, check and fix all warning-level or higher issues reported when you run tasks or commands for:
   - [ ] linting code and markdown
   - [ ] building the code
   - [ ] running all tests

> **IMPORTANT**: Fix all issues encountered, regardless of scope or relevance to changes made.
> We maintain a zero-issue code quality cleanliness.

</details>

PR #2183, opened 2026-08-10
feat: Chart Heikin-Ashi and SMA error analysis on docs pages

DaveSkender:
Wires the last two chartable doc pages from facioquo/stock-charts#498, on top of `@facioquo/indy-charts` **0.9.1**.

## Changes

- **`@facioquo/indy-charts` ^0.8.0 → ^0.9.1** — adds the candle-rendered indicator support these pages need, plus the fix in facioquo/stock-charts#542 (see below)
- **`/indicators/heikin-ashi`** — candle-rendered overlay that *replaces* the raw price candles while displayed. Both series sit at nearly the same prices, so drawing both just occludes them; volume is on its own axis and is unaffected.
- **`/indicators/sma-analysis`** — charts all three error metrics as separate panes (MAD / MAPE / MSE), since the units are incompatible (price, price², fraction). The first pane carries the price overlay for context.
- Registry entries for the four new uiids; catalog fixture refreshed from the live API (92 → 96 entries)

## The Heikin-Ashi detour

Self-review caught the chart shipping blank — volume bars and nothing else. Root cause was upstream in indy-charts 0.9.0, not here: `FinancialController.getMinMax()` folded the deliberate all-NaN padding bars into `Math.min`/`Math.max` with no finite check, so the y-scale collapsed to `0–1` and every candle clamped off-canvas. It failed worse than a missing chart, because a candle overlay also hides the raw price candles — so the page lost the price series too.

Fixed in facioquo/stock-charts#542 and released as 0.9.1. The chart was pulled in 4800c1a8 and restored in 1c66e154 once that landed.

## Verification

- `docs:build` clean
- **91/91** chart e2e tests pass (auto-discovery picked up the new pages)
- Live-API browser check on the built site, per page: canvases painted with real data (1 on heikin-ashi, 4 on sma-analysis), zero console errors, no error blocks. Confirmed by pixel-sampling the overlay canvas that the Heikin-Ashi candles paint *and* the volume band still paints, and visually that the smoothed candles replace the raw price series rather than drawing over it.


DaveSkender:
## Disposition on the swarm review

No findings to address — the review returned zero actionable items. Flagging instead that **the approval is stale**: it was submitted against b7d4705, and the PR has changed materially since.

The review checked the Heikin-Ashi wiring and cleared it, including confirming the live endpoint returns full OHLC for the new `lineType: "candle"` entry. All true — but the chart still rendered blank: only volume bars, no candles. Every registry, catalog, and API detail was correct; the defect was downstream in `@facioquo/indy-charts` 0.9.0, where `FinancialController.getMinMax()` folded the deliberate all-NaN padding bars into `Math.min`/`Math.max` without a finite check, collapsing the y-scale to `0–1` and clamping every candle off-canvas.

It failed worse than a missing chart, since a candle overlay also hides the raw price candles — so the page lost the price series too.

Changes since the approval:

- 720418a8 — keep the catalog fixture byte-faithful to the API output
- 4800c1a8 — **drop the Heikin-Ashi chart**, blocked on the upstream defect
- 3ece3fbc — ignore the local `.wrangler` state directory

The upstream fix is facioquo/stock-charts#542. Once that merges and `@facioquo/indy-charts` 0.9.1 publishes, this PR will bump to `^0.9.1` and revert 4800c1a8 to restore the chart — so a re-review will be worth running at that point.

Verified after the current changes: `docs:build` clean, 90/90 chart e2e (was 91; the heikin-ashi case is gone).

Issue #2140, opened 2026-07-05
Align result record names with indicator family names (breaking)


DaveSkender:
## Objective

Four result records do not match their indicator family name, so their files (named after the type per #1810/#2145) do not carry the indicator name. Rename the records so every indicator has a consistent `{Indicator}Result` type.

## Scope

| Indicator | Current record | Proposed | Current file |
| --------- | -------------- | -------- | ------------ |
| Correlation | `CorrResult` | `CorrelationResult` | `src/Indicators/c-d/Correlation/CorrResult.cs` |
| HtTrendline | `HtlResult` | `HtTrendlineResult` | `src/Indicators/e-j/HtTrendline/HtlResult.cs` |
| MaEnvelopes | `MaEnvelopeResult` | `MaEnvelopesResult` | `src/Indicators/k-q/MaEnvelopes/MaEnvelopeResult.cs` |
| WilliamsR | `WilliamsResult` | `WilliamsRResult` | `src/Indicators/t-z/WilliamsR/WilliamsResult.cs` |

File names follow the record renames. Includes updating all internal references, catalog listings, serialization tests, and regression baselines that key on type shape (JSON baselines key on property names, which do not change).

## Why this is breaking (and why there is no deprecation slice)

Result records are the primary public return types of the indicator APIs. A derived `[Obsolete] record CorrResult : CorrelationResult` shim does not help: consumers typed as `IReadOnlyList<CorrResult>` cannot receive the API's `IReadOnlyList<CorrelationResult>` (covariance runs the wrong direction), and record inheritance changes value-equality and `with`-expression semantics. Renaming the records is inherently a MAJOR-version change — document in the v4 migration guide instead.

## Related

- Parent: #1810
- Plan: `docs/plans/file-reorg.plan.md` (Rule 1: file name matches type name)

DaveSkender:
Confirmed still v4-only (2026-09-03), unlike the rest of the v4 bucket.

Re-checked the four records against main — `CorrResult.cs`, `HtlResult.cs`, `MaEnvelopeResult.cs`, and `WilliamsResult.cs` are all unchanged and still carry the mismatched names.

The no-deprecation-slice argument in the body holds. Result records are the primary public return types, the covariance direction is wrong for a derived `[Obsolete]` shim, and C# has no assembly-internal type alias visible to consumers (`[TypeForwardedTo]` only works across assemblies, and a `using` alias is file-local). There is no non-breaking path.

[...]

Issue #2137, opened 2026-07-05
Document the per-indicator facade class convention


DaveSkender:
> [!IMPORTANT]
> **Rewritten 2026-09-03.** The original scope rested on a misreading of the class layout and would have made naming worse. The superseded text is preserved in the issue's edit history and in the [analysis comment](https://github.com/facioquo/stock-indicators-dotnet/issues/2137#issuecomment-5532583994). This is no longer a breaking change and is no longer milestoned to v4.

## What the classes actually are

Taking EMA as the reference, `public static partial class Ema` is opened in five files:

| File | What it contributes |
| ---- | ------------------- |
| `Ema.Series.cs` | `ToEma` — Series entry point |
| `EmaList.cs` | `ToEmaList` — BufferList entry point |
| `EmaHub.cs` | `ToEmaHub` — StreamHub entry point |
| `Ema.Utilities.cs` | `Increment`, `Validate` — shared math and guards |
| `Ema.Catalog.cs` | `SeriesListing` — catalog metadata |

McGinley Dynamic has the identical shape: `MgDynamic` is opened in `DynamicHub.cs` and `DynamicList.cs` as well as its own three fragments.

So the class is **not** the series implementation. It is the per-indicator facade holding the entry point for all three styles, plus the shared utilities and the catalog listing.

## Why the original rename was wrong

Renaming `Ema` to `EmaSeries` would place `ToEmaList` and `ToEmaHub` inside a type named "Series" — less accurate than the current name, not more.

The stated goal was to match the `*Hub` / `*List` / `*Result` suffix pattern. That pattern does not apply: each of those is a concrete type for exactly one style, while the facade spans all three. The bare indicator name is the correct name for a type that is not style-specific.

## The convention, stated

- `{Indicator}` — static facade; holds `To{Indicator}`, `To{Indicator}List`, `To{Indicator}Hub`, shared utilities, and the catalog listing.
- `{Indicator}Hub`, `{Indicator}List`, `{Indicator}Result`, `I{Indicator}` — style-specific concrete types.

This is already consistent across the library. Nothing needs renaming to achieve it.

## Remaining work

- [ ] Write the convention above into the contributor documentation so the facade is not mistaken for a series class again. Natural home is alongside the existing file/type naming rules.
[...]

PR #2201, opened 2026-08-21
fix: Restore ReturnType as an obsolete alias for ResultRecordType

DaveSkender:
> ## Recommendation: merge this and release **3.1.0** — decision remains the maintainer's
>
> A structured option debate (advocate/challenger, fixed weights, fallacy check) was run
> on this fork. Verdict below; full reasoning follows. One command reverses it either way:
> merge = 3.1.0, close = 4.0.0.

## What this PR is

`IndicatorListing.ReturnType` shipped as public API in 3.0.0 (`IndicatorListing.cs:48` at the tag). #2190 renamed it to `ResultRecordType`, now on `main`. That rename is source-breaking on a strict reading: the value was always `null` in every shipped release, so nothing can depend on the value — but code that merely referenced the name compiled against 3.0.0 and no longer compiles.

This PR restores the old name as a forwarding alias:

```csharp
[JsonIgnore]
[Obsolete("Renamed to ResultRecordType. This alias is removed in the next major version.")]
public string? ReturnType
{
    get => ResultRecordType;
    init => ResultRecordType = value;
}
```

The rename is **not** undone — `ResultRecordType` stays the real, documented property. The alias forwards rather than storing (the two cannot diverge), accepts `init` (3.0.0 object-initializer code compiles too), and is `[JsonIgnore]`d (the JSON carries the value once, under the current name).

## The debate

**Question this decides:** does the pending release ship as 3.1.0 (minor, with this alias) or 4.0.0 (major, rename bare)?

**Option set** (decomposed before framing): keep both names → **A: 3.1.0 with alias** (this PR) · new name only → **B: 4.0.0 bare** (close this PR) · old name only → **C: revert the rename** — struck: the rename was maintainer-approved in #2190 and reverting reinstates the documented misnomer with no compatibility gain over A.

**Weights fixed before arguments:** consumer compatibility **High** · release describability/community signal **High** (the operator's stated constraint: one well-formed release, no wrong signals) · API cleanliness Medium · house-policy alignment Medium · mechanism upkeep Low.

### Case for A (alias, 3.1.0) — winner

- **House precedent is explicit.** `docs/migration/v3.md` documents this exact pattern as this library's way of shipping renames: *"The old `Quote`, `IQuote`, … names remain as warning-level `[Obsolete]` aliases of the new types, so existing code keeps compiling and running while deprecation warnings guide each rename."* An alias here is policy-consistent; a bare rename is the deviation.
- **With the alias, the release is backward-compatible, so semver no longer permits calling it a major.** SemVer 2.0 §7 lists deprecating existing functionality as a *minor*-version event. The only breakage the bare rename causes is source references to a property that was `null` on 243/243 listings in every shipped version (measured, this session) — and the alias removes 100% of that breakage for ~15 lines.
- **4.0.0 now collides with the queued breaking wave.** #2137–#2140 are open, explicitly-breaking rename issues. Spending the major on a never-usable property either strands that wave at 5.0.0 or forces it into this release — the exact outcome the one-well-formed-release constraint exists to prevent.

### Case for B (bare rename, 4.0.0) — discarded

Steel-manned: the semver contract is about the declared surface, not usage statistics — a public member disappeared, and the honest signal is a major; version numbers are cheap, broken trust is not. Aliases also linger: the v3 migration doc itself still documents shims that were slated for removal "in version 3.1," showing deprecation windows outlive their plans.

**Why it loses:** its core argument assumes the incompatibility it needs to prove (begging the question — with the alias merged, no incompatible change ships, so SemVer §8 never triggers). And "numbers are cheap" is false in practice: a major gates Dependabot policies, signals migration work to every consumer, and burns the number the #2137–#2140 wave needs. Scored under the fixed weights: **A 50 / B 31**.

**Cross-examination concession (A must address):** the alias fixes *source* compatibility only. The catalog JSON export key changed `returnType` → `resultRecordType` when #2190 merged, and `[JsonIgnore]` deliberately does not resurrect the old key. The old key was constitutionally `null`, so no data is lost — but a consumer validating the export against a strict JSON schema will need a one-line schema update either way.

**Fallacy/bias check (run, findings named):** B's SemVer-§8 appeal begged the question and was restated as the linger-risk argument; A's "majors scare consumers" was rejected as loss-aversion until grounded in the concrete Dependabot/wave-collision facts; the PR already existing was given **zero** weight (sunk cost); C was enumerated rather than silently dropped (false-dilemma check).

## Guard-rails from the losing side (owed on merge)

1. **Release note line:** "Catalog JSON export: `returnType` (always null) is now `resultRecordType`" — for strict-schema consumers.
2. **Record the alias's removal in the #2137–#2140 breaking wave** so the deprecation window has an end date — the v3 shims show lingering is a real failure mode, not a hypothetical.

## Verification

Alias tracks `ResultRecordType` across all 243 listings; `returnType` absent from and `resultRecordType` present in the serialized catalog — both asserted by tests. Rebased onto `c3665278`; build 0 warnings (`--no-incremental`, net10/9/8); 2525 + 76 + 4 tests passing; format and markdownlint clean.


docs/utilities/catalog.md
```diff
@@ -69,6 +69,7 @@ Each `IndicatorListing` describes one indicator-and-style combination. This is t
 | `Results` | `IReadOnlyList<IndicatorResult>` | Output fields the indicator produces |
 | `ResultRecordType` | `string?` | Name of the result record the method returns (e.g. `"EmaResult"`), the same for all three styles — the record type, not the method's literal return type (`EmaList` for Buffer, `EmaHub` for Stream). `null` when `MethodName` is unset, or when the method cannot be resolved (possible under trimming or NativeAOT publishing) — check before use |
 | `MethodName` | `string?` | Method name, for automation use cases |
+| ~~`ReturnType`~~ | `string?` | **Obsolete.** Renamed to `ResultRecordType`; forwards to it and is removed in the next major version. Not serialized. |
```
DaveSkender: Just delete this. I'm okay with some minor breaking changes and certainly don't want changelog writing style in our user facing documentation that should reflect current state and not discuss the absence of things

DaveSkender: Deleted in `d485dbb` — the row is gone entirely.

Agreed on the principle: the docs page describes the current surface, and the deprecation story already lives where a consumer actually meets it — the `[Obsolete]` message and the property's XML doc. Nothing else on the page discusses absent members.

Applied as a commit on top of your merge of `main` rather than a rewrite, so your branch history is untouched. Resolving.

